PR #24, opened 2019-02-06
Fix beqz and bgez instructions

kylelaker:
Currently beqz is not recognized in the editor as a valid command and
bgez results in an error assembling the file. There is also a begz
instruction that appears in these files which is not a valid RISC-V
instruction. This resolves that by ensuring beqz and bgez exist in all
the appropriate locations and that bgez is removed.


mortbopet:
Hi!
Thanks for finding the issue and raising the pull request - support and contributions are greatly appreciated. Issues like this (unfortunately) still pops up from time to time.

PR #29, opened 2019-02-27
Adjust splitter regex to allow parentheses

### LINKED ISSUE #10: Allow for '(' and ')' characters to exist in a label declaration
Allowing for this would make us not have to remove the function argument type when inserting RISC-V compiled code from compiler explorer.

The issue lies with the default splitter regex that is used in syntaxhighlighter.cpp in checkSyntax()

kylelaker:
This regex is a bit of a kludge and has several non-capturing groups as well as some lookaheads. It also wouldn't accept the floating point registers (but my understanding is that support for RV32F/RV32D is not available yet).

If a type has the same name as a register, then this will break. I am unsure of how that can be avoided without more complex changes than to just the regex. Hopefully in Compiler Explorer examples, that wouldn't be an issue.

Resolves #10

### DIFF
--- a/src/defines.h
+++ b/src/defines.h
@@ -143,6 +143,6 @@ typedef struct {
     uint32_t word = 0;
 } Instruction;
 
-const static auto splitter = QRegularExpression(R"((\t|\(|\)))");
+const static auto splitter = QRegularExpression(R"(\t|\((?=x(?:[1-2]\d|3[0-1]|\d)|t[0-6]|a[0-7]|s(?:1[0-1]|\d)|[sgt]p|zero)|(?:x(?:[1-2]\d|3[0-1]|\d)|t[0-6]|a[0-7]|s(?:1[0-1]|\d)|[sgt]p|zero)\K\))");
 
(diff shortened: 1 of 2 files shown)


kylelaker:
It seems like some adjustment will be necessary to `lexerutils.h:splitQuotes`

mortbopet:
Is this still WIP?

kylelaker:
It is. So it works for functions with one argument but not for anything with a comma. That will require changes to `lexerUtils.h:splitQuotes` and I'm not really sure how to go about that. I have an idea that maybe parenthesis can be treated like quotes if the regex is correct but I need to play around with that a little more.

kylelaker:
Alright, I think with these fixes to `splitQuotes`, it should be set. Let me know if there is an edge case I didn't consider that this breaks.

kylelaker:
Let me know if there are any more changes you'd like to see to this in order for it to be merged

mortbopet:
Looks good to me, will merge

PR #48, opened 2020-04-04
Replace `jalr x0 x1 0` with psuedo-op `jr x1` in examples

GameDevFox:
My editor automatically removed some white-space. If you want I can strip those changes out.

If you'd rather keep the non-pseudo instructions just let me know. Just wanted to add a convenient trick I learned today :)

Thanks for the great app!


mortbopet:
Looks good to me!

Issue #54, opened 2020-04-29
Build Instructions


ClemensFMN:
I tried to build on Linux and it seems the git submodules need to be pulled in.

`git clone https://github.com/mortbopet/Ripes.git`
`git submodule init`
`git submodule update`
`cd external/VSRTL`
`git submodule init`
`git submodule update`

`cmake .`
`make`

mortbopet:
This may alternatively be performed through the clone command; 
`git clone --recursive https://github.com/mortbopet/Ripes.git`

I've added a comment to your pull request.

PR #55, opened 2020-04-29
Update Readme.md to fix Issue #54

ClemensFMN:
See https://github.com/mortbopet/Ripes/issues/54

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,10 +39,15 @@ Initially, the following dependencies must be made available:
 
 Then, Ripes may be built as a standard CMake project:
 ```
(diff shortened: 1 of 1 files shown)


mortbopet:
Hello @ClemensFMN 
I think it would be better to add the `--recursive` call to a recommended `git clone` command. This removes the need to manually `submodule init` and `submodule update` in all subproject.

ie. `git clone --recursive https://github.com/mortbopet/Ripes.git`

ClemensFMN:
Done in https://github.com/mortbopet/Ripes/pull/56

PR #68, opened 2021-01-15
[Assembler] Hint for assembler editor

NCKUMaxSnake:
Hi @mortbopet ,
Referring to your opinion, we have a little progress in implementing this issues.
We have added a parameter at the instruction definition for determining the type (Reg or Imm) in  rv32i_assembler.cpp.
Otherwise, we also add some code to represent our desired result in instruction.h.
image
image
image

### DIFF
--- a/src/assembler/instruction.h
+++ b/src/assembler/instruction.h
@@ -103,14 +103,16 @@ struct Opcode : public Field {
 };
 
 struct Reg : public Field {
-    enum class Regsd { rd, rs1, rs2 };
     /**
      * @brief Reg
      * @param tokenIndex: Index within a list of decoded instruction tokens that corresponds to the register index
      * @param range: range in instruction field containing register index value
      */
-    Reg(const ISAInfoBase* isa, unsigned tokenIndex, BitRange range, Regsd _regsd) : Field(tokenIndex), m_range(range), m_isa(isa), regsd(_regsd) {}
-    Reg(const ISAInfoBase* isa, unsigned tokenIndex, unsigned _start, unsigned _stop, Regsd _regsd)
+    Reg(const ISAInfoBase* isa, unsigned tokenIndex, BitRange range) : Field(tokenIndex), m_range(range), m_isa(isa) {}
+    Reg(const ISAInfoBase* isa, unsigned tokenIndex, unsigned _start, unsigned _stop)
+        : Field(tokenIndex), m_range({_start, _stop}), m_isa(isa) {}
+    Reg(const ISAInfoBase* isa, unsigned tokenIndex, BitRange range, QString _regsd) : Field(tokenIndex), m_range(range), m_isa(isa), regsd(_regsd) {}
+    Reg(const ISAInfoBase* isa, unsigned tokenIndex, unsigned _start, unsigned _stop, QString _regsd)
         : Field(tokenIndex), m_range({_start, _stop}), m_isa(isa), regsd(_regsd) {}
     std::optional<Assembler::Error> apply(const Assembler::TokenizedSrcLine& line, uint32_t& instruction,
                                           FieldLinkRequest&) const override {
@@ -136,7 +138,7 @@ struct Reg : public Field {
 
     const BitRange m_range;
     const ISAInfoBase* m_isa;
-    const Regsd regsd;
+    const QString regsd = "reg";
 };
 
 struct ImmPart {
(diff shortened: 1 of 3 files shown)


src/assembler/instruction.h
```diff
@@ -103,14 +103,15 @@ struct Opcode : public Field {
 };
 
 struct Reg : public Field {
+    enum class Regsd { rd, rs1, rs2 };
```
mortbopet, line 106: General comment: In the spirit of the ISA-agnostic way that the assembler is written, it might be better for the register "names" to simply be defined as strings.

src/assembler/instruction.h
```diff
@@ -103,14 +103,15 @@ struct Opcode : public Field {
...
     /**
      * @brief Reg
      * @param tokenIndex: Index within a list of decoded instruction tokens that corresponds to the register index
      * @param range: range in instruction field containing register index value
      */
-    Reg(const ISAInfoBase* isa, unsigned tokenIndex, BitRange range) : Field(tokenIndex), m_range(range), m_isa(isa) {}
-    Reg(const ISAInfoBase* isa, unsigned tokenIndex, unsigned _start, unsigned _stop)
-        : Field(tokenIndex), m_range({_start, _stop}), m_isa(isa) {}
+    Reg(const ISAInfoBase* isa, unsigned tokenIndex, BitRange range, Regsd _regsd) : Field(tokenIndex), m_range(range), m_isa(isa), regsd(_regsd) {}
```
mortbopet, line 112: As above, regsd should be a string. Also, there could be a default argument here, for instance "r" or "reg"

src/assembler/instruction.h
```diff
@@ -103,14 +103,15 @@ struct Opcode : public Field {
...
+    Reg(const ISAInfoBase* isa, unsigned tokenIndex, unsigned _start, unsigned _stop, Regsd _regsd)
```
mortbopet, line 113: as above

src/assembler/instruction.h
```diff
@@ -289,14 +291,51 @@ class Instruction {
     }
 
     AssembleRes assemble(const Assembler::TokenizedSrcLine& line) const {
+        QString Hint = "";
+        // if(dynamic_cast<Imm*>(m_fields[1].get()))
```
mortbopet, line 295: Please don't leave commented out blocks of code in pull requests

src/assembler/instruction.h
```diff
@@ -289,14 +291,51 @@ class Instruction {
...
+                if(regField->regsd == Reg::Regsd::rs1)
+                    Hint = Hint + " [rs1]";
+                else if(regField->regsd == Reg::Regsd::rs2)
+                    Hint = Hint + " [rs2]";
+                else
+                    Hint = Hint + " [rd]";
+            }    
+        }
+
+        // Reg* regField = dynamic_cast<Reg*>(m_fields[0].get());
```
mortbopet, line 320: As above

src/assembler/instruction.h
```diff
@@ -289,14 +291,51 @@ class Instruction {
...
+        //     std::cout << "Nothing" << std::endl;
+
+        for(int i = 0; i < (m_expectedTokens-1); i++){
+            //Imm
+            if(dynamic_cast<Imm*>(m_fields[i].get()))
+                Hint = Hint + " [Imm]";
+            //Reg
+            else
+            {
+                Reg* regField = dynamic_cast<Reg*>(m_fields[i].get());
```
mortbopet, line 309: This "else" branch could be transformed to:
```
else if (auto* regField = dynamic_cast<Reg*>(m_fields[i].get())) {
    ...
}
```

src/assembler/instruction.h
```diff
@@ -289,14 +291,51 @@ class Instruction {
...
+        // if(dynamic_cast<Imm*>(m_fields[1].get()))
+        //     std::cout << "Its Imm" << std::endl;
+        // else if(dynamic_cast<Reg*>(m_fields[1].get()))
+        //     std::cout << "Its Reg" << std::endl;
+        // else
+        //     std::cout << "Nothing" << std::endl;
+
+        for(int i = 0; i < (m_expectedTokens-1); i++){
+            //Imm
+            if(dynamic_cast<Imm*>(m_fields[i].get()))
```
mortbopet, line 304: This if branch can be transformed to
```
if (auto* immField = dynamic_cast<Imm*>(m_fields[i].get())) {
    ...
}
```
And from this, i think you should also include the allowed width of the immediate value, as in your previous proposal.

src/assembler/instruction.h
```diff
@@ -289,14 +291,51 @@ class Instruction {
...
     AssembleRes assemble(const Assembler::TokenizedSrcLine& line) const {
+        QString Hint = "";
+        // if(dynamic_cast<Imm*>(m_fields[1].get()))
+        //     std::cout << "Its Imm" << std::endl;
+        // else if(dynamic_cast<Reg*>(m_fields[1].get()))
+        //     std::cout << "Its Reg" << std::endl;
+        // else
+        //     std::cout << "Nothing" << std::endl;
+
+        for(int i = 0; i < (m_expectedTokens-1); i++){
```
mortbopet, line 302: The reason why you're having to do `i < (m_expectedTokens-1)` here is because `m_expectedtokens` counts the number of tokens excluding the opcode.
In any case, i'd suggest you to drop the old-style for loop and use something like this instead
`for(const auto& field : m_fields) { ...`

mortbopet (changes requested):

PR #71, opened 2021-03-16
Added RISC-V 5 stage Processor w/o Forwarding Unit

tobias1610:
Added the functionality to select a RISC-V model w/o forward unit. Added new Checks in the hazard unit and extended it. New signals added:
- reg_do_write from the ex stage
- reg_wr_idx from the mem stage
- alu_op2_ctrl from the ex stage (For check if rd2 is immediate)
- alu_op2_ctrl from the mem stage (For check if rd2 is immediate)

Edited the layouts of the processor to look good.
Included the Fix issue when reversing across exit ECALLs.

Until now it works with a few examples. More examples are tested in the near future.


mortbopet:
Nice work! i'll get to reviewing the code as soon as possible.

The processor model is intended to not depend on user-scheduled code, correct? (ie. be able to execute any RISC-V program). If so, you can add it to the test suite, alongside the single-cycle and 5-stage models [here](https://github.com/mortbopet/Ripes/blob/assembler/test/tst_riscv.cpp#L90). That test suite will run most of the self-checking tests in [this folder](https://github.com/mortbopet/Ripes/tree/assembler/test/riscv-tests). If you come up with new tests, don't hesitate to create a pull request for those as well.

You can enable RISC-V tests by setting the `RIPES_BUILD_TESTS` variable in CMake, and afterwards the  `RIPES_ENABLE_RISCV_TESTS`. Currently, CMake checks whether there is a rv32i toolchain available in path, and CMake will fail if it doesn't find such a toolchain. This is **deprecated** since I've moved the RISC-V tests to exclusively use the internal assembler. I'll try to push a fix for this ASAP, since the whole test infrastructure needs a bit of a cleanup.

A quick-fix for this, if you don't have a toolchain available, is to move [this line](https://github.com/mortbopet/Ripes/blob/assembler/test/CMakeLists.txt#L40) alongside the other tests (line 44/45).

tobias1610:
Yes, the processor model is not depended on user-scheduled code an should be execute any RISC-V program correctly.
I've added your changes to the branch new_riscv_processor.

mortbopet:
Just checked locally - the layouts look good,
The model does not pass the self-checking tests (added the processor to `tst_riscv.h`), ie.:
```
QINFO  : tst_RISCV::testRV5_NO_FWPipeline() Running test:  "add.s"
FAIL!  : tst_RISCV::testRV5_NO_FWPipeline() Test: 'add.s' failed: Internal test error.
	 test number: 3
add.s
Register dump:	 PC:1f8
	x0:zero:	0
	x1:ra:	1
	x2:sp:	1
	x3:gp:	3
	...
```
meaning that the `add.s` test failed at [this point](https://github.com/mortbopet/Ripes/blob/master/test/riscv-tests/add.s#L16).

tobias1610:
OK thanks. Already found the bug and am working on a solution. I also noticed another mistake with jumps. How can I run the tests, I haven't found how to do it. I just set the variable  RIPES_BUILD_TESTS to ON.

mortbopet:
If you're using QtCreator, the tests should be available as a run target, as in the below image:
targets

Else, the tests should also be available in your build folder, ie. `${build folder}/test/tst_riscv`

tobias1610:
So, I fixed the error with the test. They all work fine now. Reworked the signals of the hazard unit and removed duplicate code.

But there are remaining at least one bug. If a JALR is executed after a LW instruction, the jump is not taken.

mortbopet:
> So, I fixed the error with the test. They all work fine now. Reworked the signals of the hazard unit and removed duplicate code.
> 
Great! A good final check is to ensure that all of the larger examples (ie. RanPi) works as well.

> But there are remaining at least one bug. If a JALR is executed after a LW instruction, the jump is not taken.

There unfortunately aren't any tests yet covering jump instructions properly. Please do not hesitate to convert the code you have that triggered the error into a test file in [riscv-tests](https://github.com/mortbopet/Ripes/tree/master/test/riscv-tests).

tobias1610:
I fixed some issues. The assembler examples now all work correct. But the example RanPi actual doesnt work.

mortbopet:
Hi @tobias1610, have you had any progress on getting the RanPi example to work?

tobias1610:
No, didn't make the RanPi example work properly. Returns an incorrect value. I tested the individual command categories (arithmetic, data transfer, logical, shift, condtional branch, unconditional branch, multiplication / division) by selecting individual commands from them and testing them with dependencies. No errors occurred. I have no idea where the error could be.

mortbopet:
17d5051 added cosimulation.

You should now be able to add your processor model, [as with the RV5S model](https://github.com/mortbopet/Ripes/blob/master/test/tst_cosimulate.cpp#L66), and then run the test.

When a discrepancy is detected, you will get an error output like the following:
```
FAIL!  : tst_Cosimulate::...() 
Discrepancy at cycle 103 while executing test: /../../examples/ELF/RanPi
Processor PC is: 0x128ec
Difference in register #: 16
	 expected: 0x0	actual: 0x1f
```
You can then go into Ripes (ui), place a breakpoint at the given address, and step back and forth to figure out why the discrepancy occurred.

tobias1610:
Ok, thats great. In the next days I will test it and try to fix the problem.
Sorry that my response is so late, but I've had exams for the last month and have been fully focused on them.

tobias1610:
One Question, on my local qtcreator i can't build the project. Get errors in the bool eq(.., ..) function, that 
- "class template argument deduction failed"
- "error: no matching function for call to ‘QSet(QList<QString>::const_iterator, QList<QString>::const_iterator)’"
- "error: expression list treated as compound expression in functional cast [-fpermissive]" 

Any idea why this errors occur?
Thanks for the help.

mortbopet:
Given that CI seems to pass, it might be an issue with the C++ version that you are building with.
With build errors resulting from template errors, you should ideally be able to locate the site in your program which caused the issue, as in the following image:
example

tobias1610:
I'm checked the C++ and GCC version. Both are up to date (C++17/20 and GCC 9.3.0). The error occur in the following file:
Ripes/src/isa/isainfo.h:69: error: class template argument deduction failed:
const auto ext1 = QSet(this->enabledExtensions().begin(), this->enabledExtensions().end());
error: no matching function for call to ‘QSet(QList<QString>::const_iterator, QList<QString>::const_iterator)’

I haven't changed this file or any related. I'm currently checking the Qt-Version to find out if it's because of that.

mortbopet:
It might be your Qt version - seems like this method of QSet construction was introduced in Qt 5.14: https://doc.qt.io/qt-5/qset.html#QSet-2

tobias1610:
So, I fixed the cosimulate error. The test runs without errors. The processor model should now work correctly.

Issue #127, opened 2021-10-02
[RVC] Add support for variable length instructions in the InstructionModel


mortbopet:
Currently, the InstructionModel (bottom right table in the processor tab) is built around the notion that all instructions in a program is equal in width. An example is the following, where we assume that an instruction is 4 bytes wide:
https://github.com/mortbopet/Ripes/blob/9a708c2bd83a6d587e45ed644269308881bdfc65/src/instructionmodel.cpp#L7-L12

Instead, we need a more advanced solution which informs the instruction model about relative addresses based on the result from disassembling. As a reminder, a disassembly result contains both the disassembled string as well as the number of bytes disassembled (instruction width): https://github.com/mortbopet/Ripes/blob/master/src/assembler/assembler.h#L73.

Currently we're simply taking whatever result we get from disassembling and setting that as the value in the column containing disassembled instructions, without it interacting with the rest of the instruction model logic:
https://github.com/mortbopet/Ripes/blob/9a708c2bd83a6d587e45ed644269308881bdfc65/src/instructionmodel.cpp#L132-L134

lcgamboa:
Programviewer line highlighting does not currently work because of considering all instructions 32-bit .  https://github.com/mortbopet/Ripes/blob/972390b4e861d918a8b08319fd021a377af358af/src/programviewer.cpp#L215

The information needed to calculate the line number already exists in the map I added to the instructionmodel 
https://github.com/lcgamboa/Ripes/blob/7881089e3737c9b134f15820432ea8b53825902e/src/instructionmodel.h#L54

With the m_indexToAddress map in the instructionmodel object I believe I can't access it from the programviewer's blockForAddress method.

What do you think of putting the m_indexToAddress map in the ProcessorHandler class instead of the instructiomodel class? This would solve the blockForAddress method access problem. 

I need to use programviewer instead of intructionmodel as I'm testing the unit test code https://github.com/mortbopet/Ripes/issues/137.

PR #135, opened 2021-10-08
Add support for variable length instructions in the InstructionModel

lcgamboa:
Related to https://github.com/mortbopet/Ripes/issues/127

### DIFF
--- a/src/instructionmodel.cpp
+++ b/src/instructionmodel.cpp
@@ -51,12 +51,16 @@ void InstructionModel::updateRowCount() {
     m_indexToAddress[0] = 0;
     m_rowCount = 0;
     const unsigned instrBytes = ProcessorHandler::currentISA()->instrBytes();
-    for (int addr = 0; addr < ProcessorHandler::getCurrentProgramSize();) {
-        auto disRes = ProcessorHandler::getAssembler()->disassemble(
-            ProcessorHandler::getMemory().readMem(addr, instrBytes), ProcessorHandler::getProgram()->symbols, addr);
-        addr += disRes.bytesDisassembled;
-        m_rowCount++;
-        m_indexToAddress[m_rowCount] = addr;
+    if (auto prog_spt = ProcessorHandler::getProgram()) {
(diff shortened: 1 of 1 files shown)


lcgamboa:
Unfortunately my old netbeans 8.2 doesn't have support to integrate clang-format. 
I've tried switching to vscode before, but I still find netbeans debugging support much better.
I will use git-clang-format from now.

mortbopet:
Looks like there's some CI issues that have been exposed by this PR; i'm working on the problem, and ping you once things are fixed!

mortbopet:
Would you mind rebasing your PRs onto the head of `c_ext` and re-push these PRs? apologies for the inconvenience.

mortbopet:
Copying in the comment from the other thread:

---
great progress! My initial thought is that it seems like the implementation is RISC-V-dependent. Now, this doesn't matter in the current state of Ripes (since only RISC-V is supported). However, code should in general work in an ISA-agnostic manner, such that we don't have to rewrite the entire UI whenever we switch to processors of other ISAs. If you find that there's some functionality lacking in the [ISAInfo](https://github.com/mortbopet/Ripes/blob/master/src/isa/isainfo.h#L28) interface, then we should brainstorm on this.

As suggested in my initial post, a disassembly result contains both the disassembled string as well as the number of bytes disassembled (instruction width). I imagine that this (somehow) can help with the "index-to-address" logic.
https://github.com/mortbopet/Ripes/blob/7ec2c53e4f5813ae4ccb6620e932b3251f83f367/src/assembler/assemblerbase.h#L14-L21

lcgamboa:
In the current structure the indexToAddress method is called to pass the address to the disassembleInstr method. The disassembleInstr method does not use instruction length information. 
I can change the code to use the disassembleInstr method to get indexToAddress and make the code ISA-agnostic as you suggested. I believe that using sdt:map to store the addresses of the instructions will solve the indexToAddress method optimization problem, but it will consume more memory. Today my vacation is over and this week my little son doesn't go to school, so I'm going to work more slowly.

PR #142, opened 2021-10-16
Cache disassembled version of program

mortbopet:
This commit adds caching of the disassembling of the current program to the `Program` class. This can then be used in models which rely on some form of mapping between the notions of "disassembled program line #", "disassembled instruction address" and "diassembled instruction string".


mortbopet:
@lcgamboa this moves the `m_indexToAddress` calculation to the disassembled result, which can then be used, as shown, in the instruction model, but should equally be useful to rewrite the `objdump` method, so that that method doesn't do disassembly, but simply "adds more text" onto the already disassembled program.

PR #146, opened 2021-10-18
Support to programviewer pc highlight

lcgamboa:
This PR only uses your instructionmodel code https://github.com/mortbopet/Ripes/blob/7dbf8df60a5d456db907fe663e4dc70891e6becc/src/instructionmodel.cpp#L21 to update de  calculation of the programviewer line number. 

For the current code of your disassemble_cache branch the cursor only works with assembly examples. For the C or elf examples the cursor is not update.

cursor
.


mortbopet:
the `disassemble_cache` branch has been merged into `c_ext` so this PR can be based on top of that branch.

The issue with the cursor not updating in the instruction model has been fixed in f10ef09.

PR #148, opened 2021-10-19
Define the LINUX variable before testing it

Petross404:
This slipped - LINUX variable wasn't set so the Ripes binary was not getting installed. It now works.

Sorry for any inconvenience.


Petross404:
> Looks like a VSRTL change sneaked in as well; please make sure that you're in-sync with the revision that is pointed to by master (7c44e48) - your commit points to (65ac251), the commit before (7c44e48), see https://github.com/mortbopet/VSRTL/commits/master.

Some day I will correctly merge the updated master to the branch. I have faith in me.

I will do later in the day, thanks for the help.

PR #158, opened 2021-10-23
Implementation of  5 stages processor with compressed instructions 

lcgamboa:
This PR is the initial implementation of compressed instruction support for the 5-stage processor.
I only implemented it for the simplest type, but it is possible to implement it for others as well. 
I still don't know if it is possible to implement for the 6-stage processor of 2 issues.

Since the decoder is after a register, I created an "increment" component (rv_increment.h) that controls the increment mux. 
I don't know if that would be the best way to do it. Initially I tried to implement it with an AND gate, but with my little knowledge of VSRTL I don't know if it's possible to redirect just two bits of the instruction bus to the gate's inputs. 

The examples and tests worked without problems. 

In the standard version of the diagram the component increment is not visible, in the extend it appears as in the figure below. 
5s

What is your opinion about the solution? Can I also apply to other 5-stage processors?


lcgamboa:
I'll make the adjustments you mentioned. 

@mortbopet  do you think it's better to use "Compressed decoder" or "Compress decoder" in component label?

lcgamboa:
I followed all your suggestions and updated the diagrams.

PR #160, opened 2021-10-23
New uncompress VSRTL component

lcgamboa:
In this PR I separated the part of the RVC uncompression into a new VSRTL component (Uncompress, I think it needs a new name). The decode component is returned to the original code and the new uncompression component is connected to the memory output, as in the diagram in the figure below. 

uncompress

This way the code follows the pattern already implemented in Ripes and does not require the use of a cache. This modularization will also make it easier to use the component on more-stage processors. 

In my opinion this implementation is better than the previous one. But I didn't like the result on the diagram very much. @mortbopet  in your opinion what should be done on the diagram? Or did you like it that way?

### DIFF
--- a/src/processors/RISC-V/rvss/rv_decodeRVC.h
+++ b/src/processors/RISC-V/rvss/rv_decodeRVC.h
@@ -1,0 +1,46 @@
+﻿#pragma once
+
+#include "../riscv.h"
+#include "../rv_decode.h"
+#include "../rv_uncompress.h"
+#include "VSRTL/core/vsrtl_component.h"
+
+namespace vsrtl {
+namespace core {
+using namespace Ripes;
+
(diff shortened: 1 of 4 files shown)


lcgamboa:
In that book you quoted all diagrams use increment of 4 on PC. The book only comments about compression. 
As it is new and used more for embedded processors, I believe it is not documented in architecture books anytime soon. 
From the figures I found, when the processor is single-stage, the compressed decoder and decoder are a single block. When there are more stages, they are separate blocks. I think this would be a good solution.

lcgamboa:
I created a new component rv_decodeRVC which is the union of uncompress and decode components, making the diagram go back to the same previous one. I think it's ok for the merge, feel free to suggest rename components or files.

PR #167, opened 2021-10-30
fix address map to put breakpoints

lcgamboa:
This PR fix address map to put breakpoints correctly in programviewer when RVC instructions are present.


mortbopet:
LGTM!

PR #169, opened 2021-11-11
External io tcp support

lcgamboa:
Initial release of IPC support. I only tested on Linux. After unraveling the obscure MSVC compiler error messages the code compiles in github actions. But I don't have MSVC available to test here.  
The [server](https://github.com/lcgamboa/Ripes/tree/external_io_tcp_server/TcpServer) for testing is on another branch, I don't think it's interesting to add it to the main Ripes version.


mortbopet:
> The server for testing is on another branch, I don't think it's interesting to add it to the main Ripes version.

I think since it seems like we're going to go with a non-Qt solution for the tcp client, i think it would be good to have proper tests of the solution so we can verify it on all platforms. I imagine a new `tst_xtcpsocket.cpp` test which could use this as a dummy server.

src/io/VBus.h
```diff
@@ -0,0 +1,16 @@
+
```
mortbopet, line 1: unneeded newline

src/io/XTcpSocket.cpp
```diff
@@ -0,0 +1,124 @@
+#include "XTcpSocket.h"
+
+// system headers dependent
```
mortbopet, line 3: platform dependent system headers

src/io/XTcpSocket.cpp
```diff
@@ -0,0 +1,124 @@
...
+#else
+#include <arpa/inet.h>
+#include <netinet/in.h>
+#include <sys/socket.h>
+#include <sys/types.h>
+#include <sys/un.h>
+#include <unistd.h>
+#endif
+
+// system headers independent
```
mortbopet, line 19: platform independent system headers

src/io/XTcpSocket.cpp
```diff
@@ -0,0 +1,124 @@
...
+#include <errno.h>
+#include <stdarg.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+typedef struct sockaddr sockaddr;
```
mortbopet, line 26: In C++ it's generally good style to place type definition local to a .cpp file inside an unnamed namespace. Additionally, `using` would be more appropriate, :
```c++
namespace {
    using sockaddr = struct sockaddr;
}
```

src/io/XTcpSocket.cpp
```diff
@@ -0,0 +1,124 @@
...
+#else
+    ::close(sockfd);
+#endif
+    sockfd = -1;
+}
+
+int XTcpSocket::write(const char* buff, const int size) {
+    int ret;
+    // printf("writing %i\n",size);
+    if ((ret = send(sockfd, buff, size, MSG_NOSIGNAL)) != size) {
```
mortbopet, line 49: since ret is used below;
```c+++
int ret = send(sockfd, buff, size, MSG_NOSIGNAL);
if(ret != size) {
...
```

src/io/XTcpSocket.cpp
```diff
@@ -0,0 +1,124 @@
...
+    closesocket(sockfd);
+#else
+    ::close(sockfd);
+#endif
+    sockfd = -1;
+}
+
+int XTcpSocket::write(const char* buff, const int size) {
+    int ret;
+    // printf("writing %i\n",size);
```
mortbopet, line 48: Please remove commented out code

src/io/XTcpSocket.cpp
```diff
@@ -0,0 +1,124 @@
...
+        printf("send error : %s \n", strerror(errno));
+#endif
+        return -1;
+    }
+    return ret;
+}
+
+int XTcpSocket::read(char* buff, int size) {
+    int ret;
+    // printf("reading %i ... ",size);
```
mortbopet, line 60: as above

src/io/XTcpSocket.cpp
```diff
@@ -0,0 +1,124 @@
...
+    // fflush(stdout);
+    if ((ret = recv(sockfd, buff, size, MSG_WAITALL)) != size) {
+#ifndef _MSC_VER
+        printf("recv error : %s \n", strerror(errno));
+#endif
+        return -1;
+    }
+    // printf("read %i of %i\n",ret, size);
```
mortbopet, line 68: as above

src/io/XTcpSocket.h
```diff
@@ -0,0 +1,16 @@
+
```
mortbopet, line 1: Missing include guards + extra whitespace.

src/io/XTcpSocket.h
```diff
@@ -0,0 +1,16 @@
...
+
+class XTcpSocket {
```
mortbopet, line 3: Since we're in a Qt world, i think it would be good to make this a Qt object with signals + slots. As a motivating usecase, a new signal could be created to emit error strings. Furthermore, i think it would be good to modify the interface to be Qt centric (this is both a blessing and a curse of using Qt as a "do-everything framework", which Qt apps tend to do/be forced to).
```
class XTcpSocket : public QObject {
    Q_OBJECT
public:
    XTcpSocket() { sockfd = -1; };
    void close();
    void abort();
    int write(const QByteArray& buff, const int size);
    int read(QByteArray& buff, int size);
    int isOpen(void) { return (sockfd >= 0); };
    int connectToHost(const QString& host, int port);
    int socketDescriptor(void) { return sockfd; };

signals:
    void error(const QString& msg);

private:
    int sockfd;
};

File: src/io/XTcpSocket.h [hunk 11]
```diff
@@ -0,0 +1,16 @@
...
+    void close();
+    void abort();
+    int write(const char* buff, const int size);
+    int read(char* buff, int size);
+    int isOpen(void) { return (sockfd >= 0); };
+    int connectToHost(const char* host, int port);
+    int socketDescriptor(void) { return sockfd; };
+
+private:
+    int sockfd;
```
mortbopet, line 15: `int sockfd = -1` and remove initializer from constructor.

src/io/ioexternalbus.cpp
```diff
@@ -4,13 +4,30 @@
 #include <QPainter>
 #include <QPen>
 
+#include <QtWidgets>
```
mortbopet, line 7: I dont think this is the right include; this include pulls in _all_ Qt widget headers, as far as i remember

src/io/ioexternalbus.cpp
```diff
@@ -4,13 +4,30 @@
...
+
 #include "STLExtras.h"
 #include "ioregistry.h"
 
+#ifndef _MSC_VER
+#include <arpa/inet.h>
+#include <netinet/in.h>
+#endif
+
+#define dprintf \
```
mortbopet, line 17: As above, we should use some forwarding to UI instead.

src/io/ioexternalbus.h
```diff
@@ -39,11 +42,24 @@ class IOExternalBus : public IOBase {
...
+    int32_t recv_cmd(cmd_header_t* cmd_header);
+    int32_t recv_payload(char* buff, const uint32_t payload_size);
+
     void updateAddress();
     Ui::IOExternalBus* m_ui = nullptr;
 
     std::vector<RegDesc> m_regDescs;
     std::vector<IOSymbol> m_extraSymbols;
+
+    XTcpSocket* tcpSocket = nullptr;
```
mortbopet, line 62: Memory leak on this; best option would be to make it a unique ptr:
`std::unique_ptr<XTcpSocket> tcpSocket;`

src/io/ioexternalbus.h
```diff
@@ -39,11 +42,24 @@ class IOExternalBus : public IOBase {
...
 
+private slots:
+    void connectButtonTriggered();
+    // void sktConnected();
+    // void sktRead();
+
 private:
+    uint32_t ByteSize;
+    int32_t send_cmd(const uint32_t cmd, const uint32_t payload_size = 0, const char* payload = nullptr);
+    int32_t recv_cmd(cmd_header_t* cmd_header);
```
mortbopet, line 53: Any reason why this is a pointer and not a reference?

src/io/ioexternalbus.h
```diff
@@ -39,11 +42,24 @@ class IOExternalBus : public IOBase {
...
     virtual void parameterChanged(unsigned) override;
 
+private slots:
+    void connectButtonTriggered();
+    // void sktConnected();
+    // void sktRead();
+
 private:
+    uint32_t ByteSize;
+    int32_t send_cmd(const uint32_t cmd, const uint32_t payload_size = 0, const char* payload = nullptr);
```
mortbopet, line 52: `size_t` for payload size? + payload as QByteArray? (like above)

src/io/ioexternalbus.ui
```diff
@@ -65,6 +72,9 @@
          <property name="readOnly">
           <bool>true</bool>
          </property>
+         <property name="text">
+          <string>-</string>
+         </property>
         </widget>
```
mortbopet, line 78: You could add a new read-only `QTextEdit` here for displaying messages emitted by the `XTcpSocket`.

src/io/ioexternalbus.h
```diff
@@ -39,11 +42,24 @@ class IOExternalBus : public IOBase {
 protected:
     virtual void parameterChanged(unsigned) override;
 
+private slots:
+    void connectButtonTriggered();
+    // void sktConnected();
```
mortbopet, line 47: unused code.

src/io/ioexternalbus.cpp
```diff
@@ -28,10 +45,201 @@ QString IOExternalBus::description() const {
...
+            if (send_cmd(VB_PINFO) < 0)
+                return;
+
+            cmd_header_t cmd_header;
+
+            if (recv_cmd(&cmd_header) < 0)
+                return;
+
+            if (cmd_header.payload_size) {
+                char* buff = new char[cmd_header.payload_size + 1];
```
mortbopet, line 140: In modern C++, it's generally seen as code smell to have `new`/`delete` calls in the code unless it's strictly necessary (heap memory managers etc.) - why ? since there should be some level of abstraction that handles interaction with the dynamic memory allocation system. 
Again, switching to a QByteArray here would address this issue.

src/io/ioexternalbus.cpp
```diff
@@ -28,10 +45,201 @@ QString IOExternalBus::description() const {
...
+                if (recv_payload(buff, cmd_header.payload_size) < 0)
+                    return;
+                buff[cmd_header.payload_size] = 0;
+                printf("%s\n", buff);
+
+                QJsonDocument desc = QJsonDocument::fromJson(buff);
+                // qWarning() << desc.object().value(QString("name")).toString ();
```
mortbopet, line 147: Commented out code.

src/io/ioexternalbus.cpp
```diff
@@ -28,10 +45,201 @@ QString IOExternalBus::description() const {
...
+        if (recv_cmd(&cmd_header) < 0) {
+            skt_use = 0;
+            return;
+        }
+
+        if (cmd_header.msg_type == VB_PWRITE) {
+            dprintf("write ok\n");
+        } else {
+            skt_use = 0;
+            printf("write error !!!!!!!!!!!!!!!!!!!!!!\n");
```
mortbopet, line 117: Could be sent to a ui component.

mortbopet (commented):
As i wrote above, i think having unit tests for the tcp client would be very good to ensure that this works cross-platform. A few comments on style, code cleanup and C++ norms.

src/io/VBus.h
lcgamboa: removed

test/tst_xtcpsocket.cpp
```diff
@@ -0,0 +1,62 @@
...
+        printf("Ok\n");
+    } else {
+        printf("Error\n");
+    }
+
+    memset(out, 0, 11);
+    client.write(in, 10);
+    server.read(out, 10);
+
+    if (!strcmp(in, out)) {
```
mortbopet, line 50: For this you can use the [QVERIFY](https://doc.qt.io/qt-5/qtest.html#QVERIFY) or QCOMPARE macro.

src/io/ioexternalbus.h
lcgamboa: removed

src/io/XTcpSocket.cpp
```diff
@@ -0,0 +1,124 @@
...
+#ifdef _MSC_VER
+
+static void initialize_socket(void);
+static void finalize_socket(void);
+
+struct initialize_t_ {
+    void finitialize_t_(void) { initialize_socket(); }
+};
+
+static initialize_t_ initialize_;
```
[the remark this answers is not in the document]
lcgamboa: I fixed it, MSVC doesn't support __attribute__((constructor)) like gcc, I used a static variable to work around the problem.

test/tst_xtcpsocket.cpp
lcgamboa: Yes, I did the commit to see if it would compile with MSVC. I've already made the corrections to using the QCOMPARE macro.

src/io/ioexternalbus.cpp
lcgamboa: removed

src/io/ioexternalbus.cpp
lcgamboa: I previously preferred the buff whit type char * to be able to see the values with the GDB. I changed to the QByteArray to encapsulate the dynamic memory allocation as you have indicated but now it is no longer possible to visualize the content with the GDB. Is there any way to do the GDB to view the QByteArray content?

mortbopet: Hmm, to get an equivalent output, maybe storing the `QByteArray::data()` pointer somewhere gives insight into the array.
```c++
QByteArray arr = ...
char* data = arr.data();
```
Since i mainly use QtCreator when developing, there's the added benefit of introspection into most Qt containers (https://wiki.qt.io/IDE_Debug_Helpers).

src/io/ioexternalbus.cpp
lcgamboa: Error messages are currently shown with a `QMessageBox::information` case the error occur in a different IO tab.

src/io/ioexternalbus.ui
lcgamboa: I added a spin widget to select the port number, but I could not align it, I'm not familiar with the QTCreator.

src/io/ioexternalbus.h
lcgamboa: changed following your suggestion

src/io/ioexternalbus.cpp
lcgamboa: This dprintf macro is for debugging only and will be removed in the final version.

src/io/XTcpSocket.h
lcgamboa: When I changed the XTcpSocket to inherit the QObject the tst_risc test stopped compiling due to lack of references. As I couldn't solve the problem, I didn't inherit the QObject.

src/io/ioexternalbus.h
lcgamboa: changed

mortbopet:
@lcgamboa could you elaborate a little bit on what/where crashes are triggered?

src/io/XTcpSocket.cpp
```diff
@@ -0,0 +1,193 @@
+#include "XTcpSocket.h"
+
+// platform dependent system headers
+#ifdef _MSC_VER
```
mortbopet, line 4: I would say `#ifdef MSVC` is preferred for CMake projects https://cmake.org/cmake/help/latest/variable/MSVC.html

src/io/XTcpSocket.cpp
```diff
@@ -0,0 +1,193 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+void XTcpSocket::FormatLastErrorStr(const char* func) {
+#ifdef _MSC_VER
+    int err;
+
+    char msgbuf[256];  // for a message up to 255 bytes.
+
```
mortbopet, line 30: please remove all of these extra newlines

src/io/XTcpSocket.cpp
```diff
@@ -0,0 +1,193 @@
...
+    sockfd = -1;
+}
+
+void XTcpSocket::abort() {
+    close();
+}
+
+int XTcpSocket::write(const QByteArray& buff, const size_t size) {
+    int ret = send(sockfd, buff.data(), size, MSG_NOSIGNAL);
+    if (ret != (int)size) {
```
mortbopet, line 73: please use C++ style cases (`static_cast`). This also goes for all of the `struct` casts further down.

src/io/XTcpSocket.cpp
```diff
@@ -0,0 +1,193 @@
...
+        FormatLastErrorStr("write");
+        return -1;
+    }
+    return ret;
+}
+
+int XTcpSocket::read(QByteArray& buff, const size_t size) {
+    int ret = recv(sockfd, buff.data(), size, MSG_WAITALL);
+    if (ret != (int)size) {
```
mortbopet, line 82: as above

src/io/XTcpSocket.cpp
```diff
@@ -0,0 +1,193 @@
...
+        FormatLastErrorStr("socket");
+        return 0;
+    }
+
+    memset(&serv, 0, sizeof(serv));
+    serv.sin_family = AF_INET;
+    serv.sin_addr.s_addr = inet_addr(host.toStdString().c_str());
+    serv.sin_port = htons(port);
+
+    if (connect(sockfd, (struct sockaddr*)&serv, sizeof(serv)) < 0) {
```
mortbopet, line 102: cast

src/io/XTcpSocket.cpp
```diff
@@ -0,0 +1,193 @@
...
+int XTcpSocket::serverStart(int port) {
+    struct sockaddr_in serv;
+
+    if ((listenfd = socket(PF_INET, SOCK_STREAM, 0)) < 0) {
+        FormatLastErrorStr("socket");
+        return -1;
+    }
+
+    int reuse = 1;
+    if (setsockopt(listenfd, SOL_SOCKET, SO_REUSEADDR, (const char*)&reuse, sizeof(reuse)) < 0) {
```
mortbopet, line 120: cast

src/io/XTcpSocket.cpp
```diff
@@ -0,0 +1,193 @@
...
+        FormatLastErrorStr("setsockopt(SO_REUSEADDR)");
+    }
+
+    memset(&serv, 0, sizeof(serv));
+    serv.sin_family = AF_INET;
+    serv.sin_addr.s_addr = htonl(INADDR_ANY);
+    serv.sin_port = htons(port);
+
+    if (bind(listenfd, (struct sockaddr*)&serv, sizeof(serv)) < 0) {
```
mortbopet, line 129: cast

src/io/XTcpSocket.cpp
```diff
@@ -0,0 +1,193 @@
...
+}
+
+int XTcpSocket::serverAccept(void) {
+    struct sockaddr_in cli;
+#ifdef _MSC_VER
+    int clilen = sizeof(cli);
+#else
+    unsigned int clilen = sizeof(cli);
+#endif
+    sockfd = accept(listenfd, (struct sockaddr*)&cli, &clilen);
```
mortbopet, line 151: cast

src/io/XTcpSocket.cpp
```diff
@@ -0,0 +1,193 @@
...
+        serverClose();
+        return -1;
+    }
+
+    return listenfd;
+}
+
+int XTcpSocket::serverAccept(void) {
+    struct sockaddr_in cli;
+#ifdef _MSC_VER
```
mortbopet, line 146: MSVC

src/io/XTcpSocket.cpp
```diff
@@ -0,0 +1,193 @@
...
+    int clilen = sizeof(cli);
+#else
+    unsigned int clilen = sizeof(cli);
```
mortbopet, line 149: why the difference in types for `clilen`?

src/io/XTcpSocket.cpp
```diff
@@ -0,0 +1,193 @@
...
+    sockfd = accept(listenfd, (struct sockaddr*)&cli, &clilen);
+    if (sockfd < 0) {
+        FormatLastErrorStr("accept");
+        sockfd = -1;
+    }
+    return sockfd;
+}
+
+void XTcpSocket::serverClose(void) {
+#ifdef _MSC_VER
```
mortbopet, line 160: MSVC

src/io/XTcpSocket.cpp
```diff
@@ -0,0 +1,193 @@
...
+    closesocket(listenfd);
+#else
+    ::close(listenfd);
+#endif
+    listenfd = -1;
+}
+
+int XTcpSocket::instances = 0;
```
mortbopet, line 168: For making XTcpSocket reentrant, i think it would  be good to make this an `std::atomic<int>`.

src/io/XTcpSocket.h
```diff
@@ -0,0 +1,38 @@
...
+    int serverAccept(void);
+    void serverClose(void);
+    const char* getLastErrorStr(void);
+    static int instances;
+
+    // signals:
+    //   void onError(const QString& msg);
+
+private:
+    void FormatLastErrorStr(const char* func);
```
mortbopet, line 31: Would it be possible to rewrite this to take a string ref instead of a pointer?

src/io/XTcpSocket.h
```diff
@@ -0,0 +1,38 @@
...
+    int socketDescriptor(void) { return sockfd; };
+
+    void setSocketDescriptor(int skt) { sockfd = skt; };
+    int serverStart(int port);
+    int serverAccept(void);
+    void serverClose(void);
+    const char* getLastErrorStr(void);
+    static int instances;
+
+    // signals:
```
mortbopet, line 27: unused code

src/io/XTcpSocket.h
```diff
@@ -0,0 +1,38 @@
...
+    //   void onError(const QString& msg);
+
+private:
+    void FormatLastErrorStr(const char* func);
+
+    char lastError[256];
```
mortbopet, line 33: Could also use a QString with its size reserved in the constructor.

src/io/ioexternalbus.cpp
```diff
@@ -4,13 +4,35 @@
...
 #include <QPen>
 
+#include <QJsonDocument>
+#include <QJsonObject>
+#include <QMessageBox>
+
 #include "STLExtras.h"
 #include "ioregistry.h"
 
+#ifdef _MSC_VER
```
mortbopet, line 14: msvc

src/io/ioexternalbus.cpp
```diff
@@ -28,10 +50,217 @@ QString IOExternalBus::description() const {
 }
 
 VInt IOExternalBus::ioRead(AInt offset, unsigned size) {
-    return 0;
+    uint32_t rvalue = 0;
+
+    if (tcpSocket->isOpen()) {
+        while (skt_use)
+            QApplication::processEvents();  // TODO only need in rvss, change to mutex
```
mortbopet, line 57: I see this is marked as todo, but commenting just for good measure 👍. If a component needs processor/ISA specific information, this should be retrieved through the ProcessorInterface, and never through special-case code.

src/io/ioexternalbus.cpp
```diff
@@ -28,10 +50,217 @@ QString IOExternalBus::description() const {
...
+            return 0;
+        }
+
+        if (cmd_header.payload_size) {
+            dp = QByteArray(8, 0);
+            if (recv_payload(dp, 8) < 0) {
+                skt_use = 0;
+                return 0;
+            }
+            uint32_t* payload = (uint32_t*)dp.data();
```
mortbopet, line 79: cast

src/io/ioexternalbus.cpp
```diff
@@ -28,10 +50,217 @@ QString IOExternalBus::description() const {
...
+
+            for (uint32_t i = 0; i < 2; i++) {
+                payload[i] = ntohl(payload[i]);
+            }
+            dprintf("Read addr[%x] = %x \n", payload[0], payload[1]);
```
mortbopet, line 84: Assuming these prints will be removed or changed to status message that can be emitted to the GUI before committing, so will not comment on the remainder of them.

src/io/ioexternalbus.cpp
```diff
@@ -28,10 +50,217 @@ QString IOExternalBus::description() const {
...
+void IOExternalBus::ioWrite(AInt offset, VInt value, unsigned size) {
+    dprintf("===============>ioWrite off=%lx size=%lx  value=%x\n", offset, size, value);
+    if (tcpSocket->isOpen()) {
+        while (skt_use)
+            QApplication::processEvents();  // TODO only need in rvss, change to mutex
+        skt_use = 1;
+        uint32_t payload[2];
+        payload[0] = htonl(offset);
+        payload[1] = htonl(value);
+        QByteArray dp = QByteArray(static_cast<char*>((void*)payload), 8);
```
mortbopet, line 108: casts. For this style of cast, you can use [reinterpret cast](https://en.cppreference.com/w/cpp/language/reinterpret_cast).

src/io/ioexternalbus.cpp
```diff
@@ -28,10 +50,217 @@ QString IOExternalBus::description() const {
...
+                    ByteSize = addrw * osymbols.count();
+
+                    m_ui->connectButton->setText("Disconnect");
+                    m_ui->status->setText("Connected");
+                    m_ui->server->setText(desc.object().value(QString("name")).toString());
+
+                    emit regMapChanged();
+                    emit sizeChanged();
+
+                    // delete[] buff;
```
mortbopet, line 175: unused code

src/io/ioexternalbus.cpp
```diff
@@ -28,10 +50,217 @@ QString IOExternalBus::description() const {
...
+                    }
+
+                    ByteSize = addrw * osymbols.count();
+
+                    m_ui->connectButton->setText("Disconnect");
+                    m_ui->status->setText("Connected");
+                    m_ui->server->setText(desc.object().value(QString("name")).toString());
+
+                    emit regMapChanged();
+                    emit sizeChanged();
```
mortbopet, line 173: Can these be moved to the end of the function?

src/io/ioexternalbus.cpp
```diff
@@ -28,10 +50,217 @@ QString IOExternalBus::description() const {
...
+        m_ui->status->setText("Disconnected");
+        m_ui->server->setText("-");
+
+        send_cmd(VB_QUIT);
+        tcpSocket->abort();
+
+        m_regDescs.clear();
+
+        emit regMapChanged();
+        emit sizeChanged();
```
mortbopet, line 200: As above, could be moved to the end of the function and always emitted.

src/io/ioexternalbus.cpp
```diff
@@ -28,10 +50,217 @@ QString IOExternalBus::description() const {
...
+                    m_regDescs.clear();
+                    for (QJsonObject::iterator i = osymbols.begin(); i != osymbols.end(); i++) {
+                        m_regDescs.push_back(
+                            RegDesc{i.key(), RegDesc::RW::RW, addrw * 8, static_cast<AInt>(i.value().toInt()), true});
+                    }
+
+                    ByteSize = addrw * osymbols.count();
+
+                    m_ui->connectButton->setText("Disconnect");
+                    m_ui->status->setText("Connected");
```
mortbopet, line 169: This pattern appears three times in the function. I think it could be good to have a local variable `ConnectionStatus m_status` with a `ConnectionStatus` enum. Then, we'd have a function `updateConnectionStatus()` that `switch(m_status)...` sets the UI components to reflect what is going on.

This would also allow you to remove these `m_ui` calls from here, replace them with a single assignment of the current status, and at the end of this function, a call to `updateConnectionStatus`.

src/io/ioexternalbus.cpp
```diff
@@ -28,10 +50,217 @@ QString IOExternalBus::description() const {
...
+}
+
+int32_t IOExternalBus::send_cmd(const uint32_t cmd, const uint32_t payload_size, const QByteArray* payload) {
+    int32_t ret = -1;
+    cmd_header_t cmd_header;
+
+    cmd_header.msg_type = htonl(cmd);
+    cmd_header.payload_size = htonl(payload_size);
+
+    QByteArray dp = QByteArray(static_cast<const char*>((void*)&cmd_header), sizeof(cmd_header_t));
```
mortbopet, line 211: cast

src/io/ioexternalbus.cpp
```diff
@@ -28,10 +50,217 @@ QString IOExternalBus::description() const {
...
+        return -1;
+    }
+
+    return ret;
 }
-void IOExternalBus::ioWrite(AInt offset, VInt value, unsigned size) {}
 
-void IOExternalBus::parameterChanged(unsigned paramId) {}
+int32_t IOExternalBus::recv_cmd(cmd_header_t& cmd_header) {
+    QByteArray dp = QByteArray(static_cast<char*>((void*)&cmd_header), sizeof(cmd_header_t));
```
mortbopet, line 226: cast

src/io/ioexternalbus.cpp
```diff
@@ -28,10 +50,217 @@ QString IOExternalBus::description() const {
...
+    int ret = tcpSocket->read(dp, sizeof(cmd_header_t));
+    if (ret < 0) {
+        disconnectOnError();
+        return -1;
+    }
+
+    cmd_header_t* hr = (cmd_header_t*)dp.data();
```
mortbopet, line 233: cast

src/io/ioexternalbus.cpp
```diff
@@ -28,10 +50,217 @@ QString IOExternalBus::description() const {
...
+
+    m_ui->connectButton->setText("Connect");
+    m_ui->status->setText("Disconnected");
+    m_ui->server->setText("-");
+    tcpSocket->abort();
+
+    m_regDescs.clear();
+
+    emit regMapChanged();
+    emit sizeChanged();
```
mortbopet, line 263: Since this appears everywhere where we update `m_ui->...` things, could be factored into the `updateConnectionStatus` function proposed above.

src/io/ioexternalbus.h
```diff
@@ -37,13 +40,25 @@ class IOExternalBus : public IOBase {
...
 protected:
-    virtual void parameterChanged(unsigned) override;
+    virtual void parameterChanged(unsigned) override{/* no parameters */};
+
+private slots:
+    void connectButtonTriggered();
 
 private:
+    uint32_t ByteSize;
+    int32_t send_cmd(const uint32_t cmd, const uint32_t payload_size = 0, const QByteArray* payload = nullptr);
```
mortbopet, line 50: Can we use `const QByteArray& payload = {}` instead?

test/tst_xtcpsocket.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+private:
+    XTcpSocket client;
+    XTcpSocket server;
+
+private slots:
+    void tst_pinpong();
+};
+
+void tst_xtcpsocket::tst_pinpong() {
```
mortbopet, line 18: Awesome that these tests are passing on all systems! 👍

mortbopet (commented):
Most of the comments are stylistic. A general comment would be to add a few more comments across the code to explain what is going on - makes reviewing, and maintaining(!) things a lot easier.

src/io/ioexternalbus.h
lcgamboa: Yes, of course.

src/io/VBus.h
```diff
@@ -0,0 +1,15 @@
+#ifndef VBUS_H
+#define VBUS_H
+
+// Virtual BUS Protocol ==================
+#include <stdint.h>
+
+typedef struct {
```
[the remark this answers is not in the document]
lcgamboa: Changed, old habits.

src/io/VBus.h
```diff
@@ -0,0 +1,15 @@
...
+    uint32_t msg_type;
+    uint32_t payload_size;
+} cmd_header_t;
+
+enum { VB_PINFO = 1, VB_PWRITE, VB_PREAD, VB_PSTATUS, VB_DMAWR, VB_DMARD, VB_QUIT, VB_LAST };
```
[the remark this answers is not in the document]
lcgamboa: Yes, of course.

src/io/XTcpSocket.cpp
lcgamboa: Ok, fixed

src/io/XTcpSocket.cpp
lcgamboa: Removed.

src/io/XTcpSocket.cpp
lcgamboa: Replaced.

src/io/ioexternalbus.cpp
lcgamboa: Replaced

src/io/ioexternalbus.cpp
lcgamboa: Replaced.

src/io/ioexternalbus.cpp
lcgamboa: Done.

src/io/XTcpSocket.cpp
lcgamboa: Replaced.

src/io/XTcpSocket.cpp
lcgamboa: Replaced.

src/io/XTcpSocket.cpp
lcgamboa: Replaced.

src/io/XTcpSocket.cpp
lcgamboa: Replaced.

src/io/XTcpSocket.cpp
lcgamboa: Replaced.

src/io/XTcpSocket.cpp
lcgamboa: Replaced.

src/io/XTcpSocket.cpp
lcgamboa: Replaced.

src/io/XTcpSocket.cpp
lcgamboa: Done. You are right.

src/io/XTcpSocket.h
lcgamboa: Removed.

src/io/XTcpSocket.h
lcgamboa: Yes, done.

src/io/XTcpSocket.h
lcgamboa: Replaced.

src/io/ioexternalbus.cpp
lcgamboa: Replaced.

src/io/ioexternalbus.cpp
lcgamboa: Replaced.

src/io/ioexternalbus.cpp
lcgamboa: Removed.

src/io/ioexternalbus.cpp
lcgamboa: Replaced.

src/io/ioexternalbus.cpp
lcgamboa: Yes, moved.

src/io/ioexternalbus.cpp
lcgamboa: Moved.

src/io/ioexternalbus.cpp
lcgamboa: Removed.

src/io/ioexternalbus.cpp
lcgamboa: Changed.

src/io/ioexternalbus.cpp
lcgamboa: Done.

src/io/XTcpSocket.cpp
lcgamboa: just a warning. Removed.

src/io/ioexternalbus.cpp
lcgamboa: Apparently only rvss tries to do reads and writes simultaneously, to use the socket, one operation has to finish to start another. I used a mutex to protect the socket and solve the problem. It is not architecture specific, but resource usage.

test/tst_xtcpsocket.cpp
lcgamboa: I believe we won't have any problems with sockets.

src/io/ioexternalbus.h
lcgamboa: changed to reference.

src/io/ioexternalbus.cpp
lcgamboa: Yes, you are right. Replaced.

lcgamboa:
I made all the modifications you requested. Sorry for the lack of comments in the source code, it really is my fault for not being used to working in a team.
As for errors, they do not occur in normal operation.
A simple way to test is to close the connection during Ripes' No-GUI simulation. 
To do this, just press the reset button on the PICSImLab board which resets the connection.
When the connection error occurs sometimes Ripes suffers a segmention fault instead of just showing the error message.

src/io/XTcpSocket.cpp
lcgamboa: Did not work. Only _MSC_VER is set as preprocessor variable: https://cmake.org/cmake/help/latest/variable/MSVC_VERSION.html#variable:MSVC_VERSION

src/io/ioexternalbus.h
```diff
@@ -37,13 +40,26 @@ class IOExternalBus : public IOBase {
...
 
 protected:
-    virtual void parameterChanged(unsigned) override;
+    virtual void parameterChanged(unsigned) override{/* no parameters */};
+
+private slots:
+    void connectButtonTriggered();
 
 private:
+    bool Connected = false;
```
mortbopet, line 49: Rename to `m_connected,m_byteSize`.

mortbopet:
> I made all the modifications you requested. Sorry for the lack of comments in the source code, it really is my fault for not being used to working in a team.

Thanks! don't worry too much about it, it's just another benefit of doing collaborative open source work that the people involved gain experience in giving, getting, and acting on feedback!

> A simple way to test is to close the connection during Ripes' No-GUI simulation. To do this, just press the reset button on the PICSImLab board which resets the connection. When the connection error occurs sometimes Ripes suffers a segmention fault instead of just showing the error message.

I will try this out once i have the chance :+1:. I would also recommend rebasing your branch to master since there was a bit of a nasty bug in the InstructionMatcher which caused some issues.

src/io/ioexternalbus.h
lcgamboa: Renamed,

lcgamboa:
Yes I have learned a lot working on Ripes. 
As for the rebase, as my branch is based on your `external_io`  branch, I believe you have to rebase it first in relation to the master.

mortbopet:
> As for the rebase, as my branch is based on your external_io branch, I believe you have to rebase it first in relation to the master.

You're right - should be done now :+1: .

src/io/XTcpSocket.cpp
```diff
@@ -0,0 +1,189 @@
...
+#endif
+}
+
+const QString XTcpSocket::getLastErrorStr(void) {
+    return lastError;
+}
+
+void XTcpSocket::close() {
+    // shutdown (sockfd,SHUT_RDWR );
+#ifdef _MSC_VER
```
mortbopet, line 55: What is the reason for reverting back to these?

lcgamboa: The [MSVC](https://cmake.org/cmake/help/latest/variable/MSVC.html?highlight=msvc#msvc) variable is only set for cmake scripts. The [_MSC_VER](https://cmake.org/cmake/help/latest/variable/MSVC_VERSION.html#variable:MSVC_VERSION) variable is set to the preprocessor, so it can be used in the source code (as is the case).

mortbopet:
Would you mind doing a rebase onto the current master branch? I think you've merged the master branch into this branch. By doing this, all changes in the master branch is also seen as changes in this PR, making it hard to review the code (i cannot discern your contributions from the additions in the master branch).

src/io/ioexternalbus.cpp
```diff
@@ -28,10 +45,198 @@ QString IOExternalBus::description() const {
...
+    if (payload_size) {
+        dp.append(payload);
+    }
+
+    if ((ret = tcpSocket->write(dp, dp.size())) < 0) {
+        disconnectOnError();
+        return -1;
+    }
+
+    return ret;
```
mortbopet, line 197: Suggestion (to clean it up a bit):
```c++
ret = tcpSocket->write(dp, dp.size());
if(ret < 0)
  disconnectOnError();

return ret; // assuming <0 indicates some error code, else: return ret < 0 ? -1 : ret;
```

src/io/ioexternalbus.cpp
```diff
@@ -28,10 +45,198 @@ QString IOExternalBus::description() const {
...
+}
+
+int32_t IOExternalBus::recv_payload(QByteArray& buff, const uint32_t payload_size) {
+    int ret = tcpSocket->read(buff, payload_size);
+    if (ret < 0) {
+        disconnectOnError();
+        return -1;
+    }
+    return ret;
```
mortbopet, line 222: as above

src/io/ioexternalbus.cpp
```diff
@@ -28,10 +45,198 @@ QString IOExternalBus::description() const {
...
+}
+
+void IOExternalBus::disconnectOnError(QString msg) {
+    skt_use.unlock();
+
+    if (msg.length() > 1) {
+        QMessageBox::information(nullptr, tr("Ripes VBus"), msg);
+    } else {
+        QMessageBox::information(nullptr, tr("Ripes VBus"), tcpSocket->getLastErrorStr());
+    }
```
mortbopet, line 232:
```c++
QMessageBox::information(nullptr, tr("Ripes VBus"), msg.length > 1 ? msg : tcpSocket->getLastErrorStr());
```

src/io/ioexternalbus.h
```diff
@@ -37,13 +40,25 @@ class IOExternalBus : public IOBase {
...
+private slots:
+    void connectButtonTriggered();
 
 private:
+    bool m_Connected = false;
+    uint32_t m_ByteSize;
+    int32_t send_cmd(const uint32_t cmd, const uint32_t payload_size = 0, const QByteArray& payload = {});
+    int32_t recv_cmd(VBUS::CmdHeader& cmd_header);
+    int32_t recv_payload(QByteArray& buff, const uint32_t payload_size);
+    void disconnectOnError(QString msg = "");
```
mortbopet, line 54: can we pass by const reference here? (`const QString&`).

mortbopet (changes requested):

src/io/ioexternalbus.cpp
```diff
@@ -28,10 +45,198 @@ QString IOExternalBus::description() const {
...
+        VBUS::CmdHeader cmd_header;
+
+        if (recv_cmd(cmd_header) < 0) {
+            skt_use.unlock();
+            return 0;
+        }
+
+        if (cmd_header.payload_size) {
+            dp = QByteArray(8, 0);
+            if (recv_payload(dp, 8) < 0) {
```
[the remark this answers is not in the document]
lcgamboa: Yes, of course.

src/io/ioexternalbus.cpp
```diff
@@ -28,10 +45,198 @@ QString IOExternalBus::description() const {
...
 
 VInt IOExternalBus::ioRead(AInt offset, unsigned size) {
-    return 0;
+    uint32_t rvalue = 0;
+
+    if (tcpSocket->isOpen()) {
+        skt_use.lock();
+        uint32_t payload = htonl(offset);
+        QByteArray dp = QByteArray(reinterpret_cast<const char*>(&payload), 4);
+        if (send_cmd(VBUS::VB_PREAD, 4, dp) < 0) {
```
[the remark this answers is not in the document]
lcgamboa: Ok.

src/io/ioexternalbus.cpp
```diff
@@ -28,10 +45,198 @@ QString IOExternalBus::description() const {
...
+            uint32_t* payload = reinterpret_cast<uint32_t*>(dp.data());
+
+            for (uint32_t i = 0; i < 2; i++) {
+                payload[i] = ntohl(payload[i]);
+            }
+            rvalue = payload[1];
+        } else {
+            disconnectOnError("read error");
+        }
+        skt_use.unlock();
```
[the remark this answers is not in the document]
lcgamboa: You're right, I made the switch.

src/io/ioexternalbus.cpp
lcgamboa: cleaned.

src/io/ioexternalbus.cpp
lcgamboa: exchange done.

src/io/ioexternalbus.h
lcgamboa: Yes.

src/io/ioexternalbus.cpp
lcgamboa: Ok.

src/io/ioexternalbus.cpp
```diff
@@ -28,10 +45,198 @@ QString IOExternalBus::description() const {
...
+                    return;
+                }
+                QJsonParseError error;
+                QJsonDocument desc = QJsonDocument::fromJson(buff.data(), &error);
+                if (error.error == QJsonParseError::NoError) {
+                    const unsigned int addrw = desc.object().value(QString("address width")).toInt();
+                    QJsonObject osymbols = desc.object().value(QString("symbols")).toObject();
+
+                    m_regDescs.clear();
+                    for (QJsonObject::iterator i = osymbols.begin(); i != osymbols.end(); i++) {
```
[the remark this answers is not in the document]
lcgamboa: I couldn't use it and I didn't find any example that uses it.
The range notation (using const auto&) returns an object of type QJsonValueRef which I was unable to convert to use.

src/io/ioexternalbus.cpp
```diff
@@ -28,10 +45,198 @@ QString IOExternalBus::description() const {
...
+            disconnectOnError("read error");
+        }
+        skt_use.unlock();
+    }
+    return rvalue;
 }
-void IOExternalBus::ioWrite(AInt offset, VInt value, unsigned size) {}
 
-void IOExternalBus::parameterChanged(unsigned paramId) {}
+void IOExternalBus::ioWrite(AInt offset, VInt value, unsigned size) {
```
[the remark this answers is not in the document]
lcgamboa: Ok.

lcgamboa:
I added support for passing simulation time to PICSimLab to synchronize simulators. Is there an easy way to be able to use the instruction counter value instead of the constants I added for testing?

https://github.com/lcgamboa/Ripes/blob/c3a4e5ca0031db5711a7cbded3e7378d6fb25320/src/io/ioexternalbus.cpp#L54

mortbopet:
@lcgamboa the [processor interface](https://github.com/mortbopet/Ripes/blob/master/src/processors/ripesvsrtlprocessor.h#L36) contains a hook to get the current cycle count, which can be used through the [processor handler](https://github.com/mortbopet/Ripes/blob/master/src/processortab.cpp#L303-L304).

lcgamboa:
@mortbopet  I ended up using the wall time instead of the instructions counter, it got better result since Ripes doesn't use a fixed simulation frequency. 

I followed your guidance and tried to use QTcpSocket instead of the XTcpSocket class I created. But I keep having deadlock problems because of Qt's sockets implementation and the ioRead and ioWrite implementation in Ripes. 
Running one cycle at a time works fine. 
Running in animation mode with an interval greater than 150ms works without problems, with smaller intervals occasionally deadlock occurs as several threads try to use the ioRead and ioWrite methods simultaneously.
In no-gui fast mode it works until you try to pause the simulation. When the pause is triggered several threads try to use the ioRead method at the same time and deadlock always occurs.

As it is necessary to send a command via TCP and wait for the response within the ioRead and ioWrite methods, the only possible way I found to wait is to use a QEventLoop.
To solve the problem, it would be necessary to manage at a higher level so that there was no access to the ioRead and ioWrite methods by several threads simultaneously or change the form of request of the ioRead and ioWrite methods to generate a trigger signal and wait for another signal response.
I don't have enough Ripes knowledge to make that kind of change and not whether the work would pay off. Maybe using XTcpSocket isn't such a bad idea for now.

If you want to take a look at the code using QTcpSocket it's in this branch https://github.com/lcgamboa/Ripes/blob/QTtcpSocket/src/io/ioexternalbus.cpp .

mortbopet:
> But I keep having deadlock problems because of Qt's sockets implementation and the ioRead and ioWrite implementation in Ripes.

Could you explain this a bit more? I'm assuming that the control flow of a simulation is:

1. code executes in ripes
2. an MMIO-mapped access is [read | written] to the external bus
3. this feeds into ioWrite/ioRead functions
4. This creates a tcp message which eventually will return - the Ripes simulation is stalled in the ioWrite/ioRead command while this is going on.
5. The TCP request finishes, and the simulation progresses.

Where does the deadlock come into play? where are the multiple threads coming from/doing, which tries to access the bus concurrently?

lcgamboa:
Apparently the problem is between step 4 and 5, the Ripes simulation is not always stalled.

Running step by step everything works, several threads access the ioRead and ioWrite methods.
```
ioRead  start threadID= QThread(0x555556124120)  simtime=  1683376517233706956
ioRead  stop  threadID= QThread(0x555556124120)  simtime=  1683376517233706956
ioRead  start threadID= QThreadPoolThread(0x5555572a6d90, name = "Thread (pooled)")  simtime=  1683376532943357397
ioRead  stop  threadID= QThreadPoolThread(0x5555572a6d90, name = "Thread (pooled)")  simtime=  1683376532943357397
ioRead  start threadID= QThreadPoolThread(0x55555675eb30, name = "Thread (pooled)")  simtime=  1683376533556835921
ioRead  stop  threadID= QThreadPoolThread(0x55555675eb30, name = "Thread (pooled)")  simtime=  1683376533556835921
ioRead  start threadID= QThreadPoolThread(0x5555572b7ab0, name = "Thread (pooled)")  simtime=  1683376535504702982
ioRead  stop  threadID= QThreadPoolThread(0x5555572b7ab0, name = "Thread (pooled)")  simtime=  1683376535504702982
ioRead  start threadID= QThread(0x555556124120)  simtime=  1683376535561063213
ioRead  stop  threadID= QThread(0x555556124120)  simtime=  1683376535561063213
```

In animation mode apparently the main thread tries to use ioWrite before the pooled thread finishes ioRead:
```
ioRead  start threadID= QThread(0x555556124120)  simtime=  1683376663470320350
ioRead  stop  threadID= QThread(0x555556124120)  simtime=  1683376663470320350
ioRead  start threadID= QThreadPoolThread(0x5555572ac290, name = "Thread (pooled)")  simtime=  1683376665338831463
ioRead  stop  threadID= QThreadPoolThread(0x5555572ac290, name = "Thread (pooled)")  simtime=  1683376665338831463
ioRead  start threadID= QThreadPoolThread(0x55555725f150, name = "Thread (pooled)")  simtime=  1683376665373060131
ioRead  stop  threadID= QThreadPoolThread(0x55555725f150, name = "Thread (pooled)")  simtime=  1683376665373060131
ioRead  start threadID= QThreadPoolThread(0x5555571802a0, name = "Thread (pooled)")  simtime=  1683376665438149355
ioRead  stop  threadID= QThreadPoolThread(0x5555571802a0, name = "Thread (pooled)")  simtime=  1683376665438149355
ioRead  start threadID= QThread(0x555556124120)  simtime=  1683376665472576233
ioRead  start threadID= QThreadPoolThread(0x555556f71830, name = "Thread (pooled)")  simtime=  1683376665478719220
ioRead  stop  threadID= QThread(0x555556124120)  simtime=  1683376665472576233
ioRead  start threadID= QThread(0x555556124120)  simtime=  1683376665487467718
ioWrite start threadID= QThreadPoolThread(0x555556f71830, name = "Thread (pooled)")  simtime=  1683376701053311357
"Unknown error"
ioRead  stop  threadID= QThread(0x555556124120)  simtime=  1683376665487467718
QIODevice::write (QTcpSocket): device not open
```

In fast simulation mode, when the pause is triggered apparently the main thread tries to access the ioRead several times.
```
ioWrite start threadID= QThreadPoolThread(0x55eef4307fe0, name = "Thread (pooled)")  simtime=  1683376957597663202
ioWrite stop  threadID= QThreadPoolThread(0x55eef4307fe0, name = "Thread (pooled)")  simtime=  1683376957597663202
ioWrite start threadID= QThreadPoolThread(0x55eef4307fe0, name = "Thread (pooled)")  simtime=  1683376957598503058
ioWrite stop  threadID= QThreadPoolThread(0x55eef4307fe0, name = "Thread (pooled)")  simtime=  1683376957598503058
ioRead  start threadID= QThreadPoolThread(0x55eef4307fe0, name = "Thread (pooled)")  simtime=  1683376957598720687
ioRead  start threadID= QThread(0x55eef328c120)  simtime=  1683376967598299545
ioRead  start threadID= QThread(0x55eef328c120)  simtime=  1683376967682604900
ioRead  stop  threadID= QThread(0x55eef328c120)  simtime=  1683376967682604900
ioRead  start threadID= QThread(0x55eef328c120)  simtime=  1683376967732737434
ioRead  stop  threadID= QThread(0x55eef328c120)  simtime=  1683376967732737434
ioRead  start threadID= QThread(0x55eef328c120)  simtime=  1683376967762770341
ioRead  stop  threadID= QThread(0x55eef328c120)  simtime=  1683376967762770341
ioRead  start threadID= QThread(0x55eef328c120)  simtime=  1683376967779625643
ioRead  stop  threadID= QThread(0x55eef328c120)  simtime=  1683376967779625643
ioRead  stop  threadID= QThread(0x55eef328c120)  simtime=  1683376967598299545
Falha de segmentação
```

To wait for the Socket's response, it is necessary to use a local QEventLoop to process the signals and events, I believe that this is why the simulation does not stalled.

Issue #181, opened 2021-11-26
Add support for autoformatting source code in Assembly or C 


lcgamboa:
> An automatic code formatting option would also be helpful. 

Source code auto formatting is a common tool in most editors. Although not simple to implement it would be a nice addition for Ripes users. 

_Originally posted by @mortbopet in https://github.com/mortbopet/Ripes/issues/162#issuecomment-978976066_

mortbopet:
Currently looking in to this. My plan is to hook into using a formatter program (like `clang-format`) and run it on file save.

PR #182, opened 2021-11-28
Add clang-format support #181

mortbopet:
This commit adds support for running clang-format on C source code. `clang-format` must be available in PATH.

### DIFF
--- a/src/ccmanager.cpp
+++ b/src/ccmanager.cpp
@@ -4,6 +4,7 @@
 #include "loaddialog.h"
 #include "processorhandler.h"
 #include "ripessettings.h"
+#include "utilities/systemutils.h"
 
 #include <QProcess>
 #include <QProgressDialog>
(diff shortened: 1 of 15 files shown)


mortbopet:
@lcgamboa do you mind trying this out before i merge it?

lcgamboa:
I have tested here on Linux on various files and it worked without any problems.
Some doubts:
If clang-format is not available what happens?
Is it possible to choose the executable path? (Or use another one formatter ?)

mortbopet:
> If clang-format is not available what happens?
Nothing will happen.

> Is it possible to choose the executable path?
At the moment no, but i can work that in. 

> (Or use another one formatter ?)
Currently, no. The issue with using a formatter integrated into an IDE is that the formatter needs to provide some form of curser mapping support (if the cursor was at position x before formatting, where should it be after formatting, so as to maintain its position in the code?). [clang-format](https://clang.llvm.org/docs/ClangFormat.html) has the `--cursor` argument which i use for this. I am unsure about other formatters, but there probably needs to be specialized handling for each formatter type. Which is why things are limited to clang-format for now.

mortbopet:
Added option to set clang-format executable as well as modified where the temporary file is being generated, such that clang-format will use any .clang-format file in the source file directory, if available.

PR #183, opened 2021-11-28
Fix bgtu and bleu pseudo-instructions to swap tokens correctly

marcom97:
The bgtu and bleu instructions were using the same token twice when generating the corresponding instruction. This was fixed by correctly swapping the first two tokens of these pseudo-instructions as expected.


mortbopet:
Thank you for catching this!

PR #195, opened 2022-02-05
Fix offset of compressed instructions c.bnez and c.beqz

lcgamboa:
This PR fixes a bug in decoding the compressed instructions c.bnez and c.beqz.

### DIFF
--- a/src/processors/RISC-V/rv_uncompress.h
+++ b/src/processors/RISC-V/rv_uncompress.h
@@ -297,7 +297,7 @@ public:
                         case 0b010: {  // c.lwsp
                             const auto fields = RVInstrParser::getParser()->decodeCI16Instr(instrValue);
                             rd = fields[3];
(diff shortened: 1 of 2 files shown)


mortbopet:
Great find! Would you mind adding a test in https://github.com/mortbopet/Ripes/blob/master/test/riscv-tests-c/rvc.s which captures the issue?

lcgamboa:
> Great find! Would you mind adding tests in https://github.com/mortbopet/Ripes/blob/master/test/riscv-tests-c/rvc.s which captures the issues?

To test these instructions, is needed to make big jumps backward and forward, to test the limits of arguments. For example c.bnez x15, 255 and c.bnez x15, -255. Do you know any way to do this without having to use a bunch of nop instructions to fill in addresses between jumps?

mortbopet:
One way is to use the `.zero` directive inside the `.text` segment:
image

PR #198, opened 2022-02-18
Add docker recipe

matsievskiysv:
Add recipe to build and run Ripes from the docker image.

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -1,0 +1,30 @@
+name: Build docker image
+on:
+  push:
+    branches:
+      - master
+  workflow_dispatch:
+
+jobs:
+  docker:
+    runs-on: ubuntu-latest
+    steps:
+      -
+        uses: actions/checkout@v2
+      -
+        name: Set up Docker Buildx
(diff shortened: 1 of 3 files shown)


matsievskiysv:
I've created a workflow for building a docker image and uploading it to DockerHub.
In order to use it you'll need
1. Create `DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN` secrets with your dockerhub credentials
1. Adjust user the last line in `.github/workflows/docker-build.yml`: substitute `matsievskiysv` with your dockerhub username

> for testing this in CI

A subset of tests (which don't need X) are executed after build so if build or tests fail, workflow would also fail.

mortbopet:
Awesome, I've created an account at dockerhub and added the necessary secrets. I'll go ahead and merge and then we'll take it from there based on whether things work.

matsievskiysv:
> and added the necessary secrets

You should also change the tag.
FYI, this workflow can be started manually form actions page if needed.

mortbopet:
Should be settled now https://hub.docker.com/r/mortbopet/ripes.
Thank you very much for the contribution!

Issue #174, opened 2021-11-19
Add counters to see the number of stalls and flushes during the code execution


lcaimi:
Issue from discussion [#173](https://github.com/mortbopet/Ripes/discussions/173).

matms:
Hello, I've written a basic implementation of this feature for rv5s. Currently the other CPUs simply report no stalls or flushes, so this is likely something we'd want to work on, possibly before merging? I was wondering if anyone could take a look at it, and tell me whether I should submit a PR: https://github.com/matms/Ripes/commit/f939da079db3c6744d28a99fa66a73665151bf8d.

(Note that I'm not actually counting flush "events", rather I simply consider each instruction flushed as a cycle "lost" due to that flush, and report _that_ instead.)

mortbopet:
Hi @matms

Thank you for looking into this. I'm wondering whether a better approach would be to use the [StageInfo](https://github.com/mortbopet/Ripes/blob/master/src/processors/interface/ripesprocessor.h#L124) data already exposed through the processor interface - [this is what is used to generate visualizations showing flushes/stalls](https://github.com/mortbopet/Ripes/blob/master/src/processortab.cpp#L415).

This could be queried every cycle to determine stalls and flushes. This then leads to the question of what defines a stalled/flushed stage and how is it uniqued (or in other words, if this interface is all we have, how to we make sure that we don't record stalls/flushes more than once?).
For now, it would probably be sufficient to just assume that whenever **the last stage** is stalled/flushed, then that increments the counter. StageCount is accessed through [this function](https://github.com/mortbopet/Ripes/blob/master/src/processors/interface/ripesprocessor.h#L98).

Let me know what you think!

matms:
I've made a quick change to now use StageInfo https://github.com/matms/Ripes/commit/b6a30c53e1c7f4fbbeeeed1bc78b98a9782f1998. 

Also, I suppose it could be a good idea to change clockProcessor() (and reverse())'s handling of m_instructionsRetired to also use stageInfo (stageValid), but I'm not sure.

mortbopet:
> I've made a quick change to now use StageInfo

[...]

PR #203, opened 2022-03-26
Stall and Flush Counter (Initial Iteration)

matms:
As per #174, I am opening a PR for adding a stall and flush counter.

However, note this PR is probably not quite mergeable yet. In particular, there are two "bugs", so we'd need to discuss how to fix these. In particular:

- An instruction is considered and counted as stalled or flushed when it reaches WB, rather than when it _leaves_ WB. This diverges from the behaviour of instructionsRetired, which counts an instruction as retired upon _leaving_ WB.

- The dual-issue processor currently considers only only one way for counting stalls and flushes. I suspect we might need a way to generically query a processor for which stages contain instructions that are about to be retired. Right now I'm using `proc->stageInfo(proc->stageCount() - 1)` to gather information about these instructions, which only works for the single ~~cycle~~ issue processors.

Once both these issues are resolved, we should be able to refactor instructionsRetired to be implemented in a manner similar to the stall and flush counter.

(Also, the diff for processortab.ui is quite messy... I hope this is not an issue)


mortbopet:
> which only works for the single cycle processors.

do you mean single issue processors?

> I suspect we might need a way to generically query a processor for which stages contain instructions that are about to be retired.

I've considered whether we need to switch up the way that a stage reports its # of stages. By only having the ability to specify # of stages, we miss out on the processor telling about its structure, as in, # of issue slots/lanes. Which then restricts itself to single-issue processors.
Instead, a processor could return a `std::map<int, int>` of specifying the lane # (key) and the stages in that lane (value). For the existing processors, this would be:
* single cycle: `{(0, 1)}`
* 5 stage: `{(0, 5)}`
* 5 stage: `{(0, 6), (1, 6)}`

StageInfo would then follow this mapping, wherein it is indexed by a tuple of `(lane, stage #)`. From there, it would then be up to the stall/flush counting mechanism to determine how to map this information to a counter. What do you think about this?

> (Also, the diff for processortab.ui is quite messy... I hope this is not an issue)
No worries, I'm assuming that you're using the UI Design tab in QtCreator, and so as long things look and act as expected, then all good!

matms:
I've updated the flush and stall counter to use the new ProcessorStructure interface.

Sorry for the delay in updating, I've been quite busy the past few weeks.

PR #226, opened 2022-09-07
Fix crash when using extended layout for 5-stage processor.

kosayoda:
Prior to this commit, switching to the extended layout for the 5-stage processor would crash Ripes with a 'std::out_of_range' exception caused by `map::at`.
This renders Ripes unusable until `processor_layout_id=1` was removed or set to `0` in the configuration as Ripes would also crash on startup.

### DIFF
--- a/src/processorregistry.cpp
+++ b/src/processorregistry.cpp
@@ -125,11 +125,11 @@ ProcessorRegistry::ProcessorRegistry() {
                {{0, 4}, QPointF{0.87, 0}}}},
              {"Extended",
               ":/layouts/RISC-V/rv5s/rv5s_extended_layout.json",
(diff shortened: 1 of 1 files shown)


mortbopet:
Thank you for reporting this - I can reproduce it locally. Will have a look to find out what's going wrong here!

kosayoda:
> Thank you for reporting this - I can reproduce it locally. Will have a look to find out what's going wrong here!

My PR should fix this issue unless I did something wrong 😅

mortbopet:
> My PR should fix this issue unless I did something wrong 😅

Woops, sorry about that - very used to people submitting issues, and not so used to PRs. Thank you very much, that LGTM!

PR #278, opened 2023-04-25
compatibilty with g++

nate-4000:
i changed like 10 bytes

### DIFF
--- a/src/isa/rv32isainfo.h
+++ b/src/isa/rv32isainfo.h
@@ -36,7 +36,7 @@ public:
 
     return march;
   }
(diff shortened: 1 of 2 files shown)


mortbopet:
Are you sure this is the case for all compilers? As far as i can tell, relevant documentation for the bare-metal toolchain(s) still reference the existing ABI strings:
https://www.sifive.com/blog/all-aboard-part-1-compiler-args
https://gcc.gnu.org/onlinedocs/gcc/RISC-V-Options.html
Are you sure that you are using a bare-metal compiler?

PR #288, opened 2023-06-26
Add missing LD_LIBRARY_PATH variable with Qt libs in ripes.dockerfile

kochaika:
Without this variable the following error occurs when running an application in a container:
`Ripes: error while loading shared libraries: libQt6Charts.so.6: cannot open shared object file: No such file or directory`


mortbopet:
Thanks! If possible, it could be cool if the docker build could also be tested in CI!

PR #289, opened 2023-07-07
Branch prediction processor

anglothefloof:
Implemented a 5-stage branch prediction processor, along with a tab to control its properties like predictor type and whatnot, along with statistics and visualizations of internal state. Once the processor with branch prediction is selected through the processor selection menu, the branch prediction tab is enabled. The processor will hold its internal state through different program changes and reloads, allowing multiple iterations of running a single program or seeing how different programs interact through the predictor. The user can reset the state at any time if they wish. It supports reversing execution, RV32IMC, and RV64IMC.

Small note: the test programs (such as tst_riscv) supplied with Ripes do not work with the processor, but I did run each test assembly file manually and they all were successful. The errors (usually segmentation faults) originate from different places at different times, sometimes consistent between runs and sometimes producing different errors. Possible race condition?

However, in the Ripes software itself, it is very stable and error free, at least as far as I've been able to test.

If there's any changes that need to be made, I'd be more than glad to do them, and any guidance on the aforementioned issue would be greatly appreciated.


mortbopet:
Awesome for taking a stab at this, can't wait to try it out! Its a big PR so let's make sure that we take our time and get it right - given the size I'll be reviewing it in stages.

mortbopet:
> however it might prove difficult to "detach" the branch predictor from the processor implementation considering that the branch predictor itself is part of the processor hardware

Oh most definitely, and there's probably no way around having to have an e.g. "RISC-V branch predictor"-gasketmodule. But that should just be a gasket which performs any intermediate conversions required to use the generic branch predictor model.

Issue #122, opened 2021-09-29
Elaborate documentation of file-related system calls


mortbopet:
As described in #121, the system call menu `help->system calls` has lacking documentation for the usage and meaning of file-related system calls, such as `open`. 

These descriptions should be elaborated, such as by including the semantics of the various flags for `open`:
https://github.com/mortbopet/Ripes/blob/d692d719e60dd9e6f043be337dbfa003a20dcba8/src/syscall/systemio.h#L80-L86

Satarupa22-SD:
@mortbopet I have added a  [PR](https://github.com/mortbopet/Ripes/pull/291) on this issue. Could you please check and let me know if any additional changes are to be made. Thanks!

PR #291, opened 2023-07-26
Update systemio.h

Satarupa22-SD:
Elaborated descriptions as per issue #122

### DIFF
--- a/src/syscall/systemio.h
+++ b/src/syscall/systemio.h
@@ -75,13 +75,31 @@ private:
   static constexpr int SYSCALL_BUFSIZE = 128;
   // Maximum number of files that can be open
   static constexpr int SYSCALL_MAXFILES = 32;
+  // Flags for open system call (following Unix conventions):
 
+  // Opens the file for reading data only. File's content can't be modified.
   static constexpr int O_RDONLY = 0x00000000;
+
+  // Opens the file for writing data only. File's content can't be read.
   static constexpr int O_WRONLY = 0x00000001;
+
+  // Opens the file for both reading and writing.
   static constexpr int O_RDWR = 0x00000002;
+
+  // Append
+  // Adds the data to the end of the file instead of overwriting it.
   static constexpr int O_APPEND = 0x00000008;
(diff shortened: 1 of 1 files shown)


Satarupa22-SD:
@mortbopet Description:
The `open` system call is used to open files in the file system. It allows you to create new files or open existing ones for reading, writing, or both. The call returns a file descriptor that represents the opened file, which is later used for performing various operations on the file.
we can also include syntax, parameters(path and flags) and modify the return value to further improve the current documentation. 

However could you please share the file location where the changes need to be made. I have not been able to find it . Thanks!

mortbopet:
2nd argument here is the description that'll appear in the dialog https://github.com/mortbopet/Ripes/blob/c36b9c76feffd2a015654956842b0e0fd26486b0/src/syscall/file.h#L17

Satarupa22-SD:
@mortbopet I have made the changes to the file.h file. Please let me know if any additional changes are to be made

mortbopet:
Great, thanks! I think the description should also include information about the different flags that the user can provide as argument - that, or just point to a link describing these since i assume it's documented somewhere in the RISC-V ABI.

Satarupa22-SD:
@mortbopet I have updated the flag description. Please check and let me know if any modifications are needed.

mortbopet:
> has the build failed due to the flag description which I have added?

Correct - make sure that the application is able to build successfully locally; looks like this is an issue with `addFlagDescription` not being declared anywhere.

Satarupa22-SD:
@mortbopet has this issue been reassigned? I was busy with University stuff so hence was off GitHub for a few days.

Satarupa22-SD:
Also when can we expect to hear about the status of our LFX mentorship application.

mortbopet:
The issue in question is that there's an issue in the code you submitted, where `addFlagDescription` is not being declared anywhere - so you need to fix that.

Satarupa22-SD:
@mortbopet I will start working on it then. Thanks!

PR #294, opened 2023-08-09
Implement persisting UI states

### LINKED ISSUE #140: Persisting UI state between program invocations
Currently, things such as the program loaded into the editor and the active processor are stored as settings and persisted between program invocations.
We should be able to add more features like this, to persist things such as:
- Window sizes and states
- register display modes
- memory display modes
- ...

tech-bash:
Fix: #140 

I have started the task, please let me know if I am going in the right direction. Just for my check the `moveEvent` doesn't have any size() function it has ``QPoint`` so to avoid the implications I set the`m_savedSize` type to ```QVariant``` would that be okay?

### DIFF
--- a/src/console.cpp
+++ b/src/console.cpp
@@ -43,6 +43,10 @@ Console::Console(QWidget *parent) : QPlainTextEdit(parent) {
           });
 
   m_localEchoEnabled = RipesSettings::value(RIPES_SETTING_CONSOLEECHO).toBool();
+
+  // Retrieve and store saved window size and state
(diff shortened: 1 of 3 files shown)


tech-bash:
@mortbopet any comments sir?

mortbopet (commented):
I don't think there's a need to store the size and state in in-memory variables. Instead, you should call `RipesSettings::setValue` within the events.
Secondly, there's two parts to UI state persistence - one is storing the current state whenever it changes, the other is that you need to load said state upon program startup (i suggest that you analyse some of the other usages of `RipesSettings` throughout the codebase to get a feel for this).

PR #296, opened 2023-09-01
Fix write syscall

JackWilli:
The fdInUse function passed in O_WRONLY | O_RDWR as flags and then checked fileFlags[fd] & flags == flags. This is only true if all of the bits set in flags are also set in fileFlags[fd]. Since O_WRONLY | O_RDWR are mutually exclusive the fdInUse function always returns false in this case and the write syscall fails.

I fixed the expression and passed in an enum instead to make it more explicit about what we are testing for.

### DIFF
--- a/src/syscall/systemio.h
+++ b/src/syscall/systemio.h
@@ -84,6 +84,11 @@ private:
   static constexpr int O_TRUNC = 0x00000400; // 1024
   static constexpr int O_EXCL = 0x00000800;  // 2048
 
+  enum Permission {
+      WRITE_PERMS,
+      READ_PERMS,
+  };
+
   // //////////////////////////////////////////////////////////////////////////////
   // Maintain information on files in use. The index to the arrays is the "file
   // descriptor."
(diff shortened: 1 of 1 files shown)


src/syscall/systemio.h
```diff
@@ -184,18 +189,16 @@ class SystemIO : public QObject {
...
         return false;
-      } else if ((fileFlags[fd] & flag) ==
-                 static_cast<unsigned>(
-                     flag) /* also compares ie. O_RDONLY (0x0) */) {
-        return true;
+      } else if (perm == READ_PERMS) {
+        return (fileFlags[fd] + 1) & 1; // LSB is 1 if the fd has read perms
       }
-      return false;
+      return (fileFlags[fd] + 1) & 2; // 2nd LSB is 1 if the fd has write perms
```
mortbopet, line 201: Mixing arithmetic and bitwise logic is a bit too funky for me. I think these expressions can be rewritten using bit extracts and XORs.

src/syscall/systemio.h
```diff
@@ -84,6 +84,11 @@ class SystemIO : public QObject {
   static constexpr int O_TRUNC = 0x00000400; // 1024
   static constexpr int O_EXCL = 0x00000800;  // 2048
 
+  enum Permission {
```
mortbopet, line 87: enum class

src/syscall/systemio.h
mortbopet: secondly, we should also extract out the _nth LSB is X perms_ to an enum, e.g. something like
```c++
enum FilePermissions {
   RD_PERM_FLAG = 1 << 0,
   WR_PERM_FLAG = 1 << 1
};
```

mortbopet (commented):
Some comments - bonus points if you can also come up with a unit test to verify your changes :).

Issue #301, opened 2023-09-22
Refactor instruction definitions from macros to classes


mortbopet:
Currently, defining instructions for the assembler is heavily dependent on macro usage (https://github.com/mortbopet/Ripes/blob/master/src/assembler/rvassembler_common.h#L12-L108):

In most cases, classes and templates can replace macros - i certainly think that could be the case here. I'd imagine that instead of having macros here, we'd have some hierarchy of classes that inherit from [Instruction](https://github.com/mortbopet/Ripes/blob/master/src/assembler/instruction.h#L391). Most likely, this would result in a class hierarchy of
> `Instruction -> RVInstruction -> [BTypeInstr, ITypeInstr, ...]`

possibly with templates sprinkled in to denote the register width that the assembler is being initialized with.

The intention of this change is to
1. Remove macros - they're hard to debug, cryptic, and easy to get wrong, and IDEs hate them.
2. Classes should allow us to share a bit more code, making adding new instructions simpler
3. ... and allow us to easily add new fields to the instructions, such as more elaborate descriptions of the instruction itself, its fields, ... - without it becoming a complete mess.

PR #302, opened 2023-09-22
Refactor instruction definitions from macros to classes

raccog:
Draft branch for #301.

### DIFF
--- a/src/assembler/rv_c_ext.h
+++ b/src/assembler/rv_c_ext.h
@@ -236,6 +236,21 @@ public:
                       std::vector{ImmPart(6, 7, 8), ImmPart(2, 9, 12)}) {}
 };
 
+// A RISC-V unsigned immediate field with an input width of 9 bits.
+// Used in C.SDSP and C.FSDSP instructions.
+//
+// It is defined as:
+//  - Imm[8:6] = Inst[9:7]
+//  - Imm[5:3] = Inst[12:10]
+//  - Imm[2:0] = 0
+template <typename Reg_T>
+class RVCImmSDSP : public RVCImm<Reg_T> {
+public:
+  RVCImmSDSP()
+      : RVCImm<Reg_T>(2, 9, Imm<Reg_T>::Repr::Unsigned,
+                      std::vector{ImmPart(6, 7, 9), ImmPart(3, 10, 12)}) {}
+};
+
 template <typename Reg_T>
 class CATypeInstr : public RVCInstruction<Reg_T> {
 public:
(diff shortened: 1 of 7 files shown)


src/assembler/rvassembler_common.h
```diff
@@ -6,49 +6,208 @@
...
+             std::make_shared<Imm<Reg_T>>(RVImm13<Reg_T>(3))
+            }) {}
+};
+
+template <typename Reg_T>
+std::shared_ptr<Instruction<Reg_T>> defineBType(const QString &name, unsigned funct3,
+                                                       const ISAInfoBase *isa) {
+  return std::shared_ptr<Instruction<Reg_T>>(
+      new BTypeInstr<Reg_T>(Token(name), funct3, isa));
+}
```
mortbopet, line 127: If the templates are this small, then we might aswell just call `std::make_shared` at the callsite, instead of these `define` functions.

src/assembler/rvassembler_common.h
```diff
@@ -6,49 +6,208 @@
...
-       std::make_shared<_Reg>(isa, 2, 20, 24, "rs2"),                          \
-       std::make_shared<_Imm>(                                                 \
-           3, 13, _Imm::Repr::Signed,                                          \
-           std::vector{ImmPart(12, 31, 31), ImmPart(11, 7, 7),                 \
-                       ImmPart(5, 25, 30), ImmPart(1, 8, 11)},                 \
-           _Imm::SymbolType::Relative)}))
+class RVOpPartOpcode : public OpPart {
+public:
+  RVOpPartOpcode(RVISA::Opcode opcode)
+      : OpPart(opcode, 0, 6) {}
```
mortbopet, line 21: Would be nice with a little note for all of these with the style of "All RISC-V opcodes are defined as the 7 LSBs of the instruction" or similar.

src/assembler/rvassembler_common.h
```diff
@@ -6,49 +6,208 @@
...
+class RVRegRd : public Reg<Reg_T> {
+public:
+  RVRegRd(const ISAInfoBase *isa, unsigned fieldIndex)
+      : Reg<Reg_T>(isa, fieldIndex, 7, 11, "rd") {}
+};
+
+template <typename Reg_T, class RVReg_T>
+std::shared_ptr<Reg<Reg_T>> makeRvReg(const ISAInfoBase *isa, unsigned fieldIndex) {
+  return std::make_shared<Reg<Reg_T>>(RVReg_T(isa, fieldIndex));
+}
```
mortbopet, line 80: A bit in a similar vein - IMO too small to warrant a separate function.

mortbopet (commented):
It's quite a few additional lines, but that were to be expected. I suspect changing these to structs (removes the public keyword) and potentially removing the `Reg_T` templating in the future, that will cut out a lot of the boilerplate in these files.

src/assembler/rv_c_ext.h
```diff
@@ -445,7 +245,23 @@ class RVCInstrCAType : public RVCInstruction<Reg_T> {
                              RVCOpPartFunct2(funct2, RVCOpPartFunct2::OFFSET5),
                              RVCOpPartFunct6(funct6)),
             {std::make_shared<RVCRegRs2Prime<Reg_T>>(isa, 2),
-             std::make_shared<RVCRegRdRs1Prime<Reg_T>>(isa, 1)}) {}
+             std::make_shared<RVCRegRdRs1Prime>(isa, 1)}) {}
+
+  /// All RV-C Funct6 opcode parts are defined as a 6-bit field in bits 10-15
+  /// of the instruction
+  class RVCOpPartFunct6 : public RVCOpPart {
```
mortbopet: Ubernit - when `class RVCOpPartFunct6` is nested within `class RVCInstrCAType`, prefixing the former with `RVC` feels redundant to me; this goes for all of the nested classes in this PR.

mortbopet (approved):
LGTM apart from the naming nit.

mortbopet:
Tip: most editors can enable clang-format on save (e.g. VSCode with https://marketplace.visualstudio.com/items?itemName=xaver.clang-format). Then you'll never run into CI complaining about it!

raccog:
I have clang-format enabled on save in Qt Creator but it wouldn't format `rv_c_ext.h` for some reason. I had to format it manually with the command lol. Anyways, should be all set now.

Issue #311, opened 2023-09-26
Investigation: Does templating the entire assembler infra on register width really matter?


mortbopet:
Currently, **all** of the assembler and related logic is templated on types that match the register width of the target architecture (e.g. `uint32_t` for 32-bit, `uint64_t` for 64-bit) - and someplace also instruction width.

This absolutely infects all of the related code, which usually is an indication that it's not a good design pattern, e.g.:
https://github.com/mortbopet/Ripes/blob/master/src/assembler/assembler.h#L61-L76

It would be interesting to see whether providing register width dynamically would only be a neglible slowdown as opposed to the current implementation. If so, I personally think it's a valid tradeoff, given that we can simplify a bunch of code and remove templates.

mortbopet:
Fixed in 2a042c9

Issue #306, opened 2023-09-25
Make ISA definitions compile-time defined.


mortbopet:
From https://github.com/mortbopet/Ripes/issues/303:

> Currrently, instructions are defined dynamically (e.g. https://github.com/mortbopet/Ripes/blob/master/src/assembler/rv_c_ext.h#L319C11-L323), meaning that all information that we specify when defining an instruction must lookup some runtime object to acquire the values of interest (opcode bitfields etc..). However, from the point of view of writing e.g. the RISC-V decoder model, we only need static information - hence there would be quite a bit of overhead in performing a bunch of dynamic, runtime lookups. As such, solving this issue should also consider whether there is a way to make instruction information (specifically bitfields) available at compile time.

So as suggested above, it could be _neat_ if we would have a way to statically define an instruction set, rather than the current dynamic, runtime-defined method.

Some prototyping work to show the feasability: https://godbolt.org/z/zjhr9nTfq

Cool things that such a compile-time defined approach would allow for:
* Deduplicating information and increasing performance - we can, at compile time, lookup any field of any instruction and perform computations with said fields
* Compile-time check that instructions are sanely defined
  * e.g. all bit fields must be accounted for
  * and there must be no bit-fields that overlap

mortbopet:
fixed in [2a042c9](https://github.com/mortbopet/Ripes/commit/2a042c91fdb496c1fd485a28a08c34c8d66f2c7f)

Issue #307, opened 2023-09-25
Extract ISA and assembler to a separate repo/project


mortbopet:
Depends on #301, #303, #306

The ISA definition and assembler/disassembler library in Ripes isn't inherently tied to Ripes. Rather, it represents a neat little library for defining relevant information of an ISA, and then based on this, automatically generate assemblers and disassemblers from said information.

Back when I initially wrote the assembler/disassembler for Ripes i looked around for any good opensource alternatives, of which (back then) i found none. Everything was either too ISA specific or sauced into some larger context.  
The benefit of the Ripes assembler/disassembler/isa infrastructure is that it's **generic** - meaning that it should ideally work with **any** instruction set. Second, it's (relatively speaking) **tiny** compared to e.g. using LLVM. As such, I think it could be a great little open-source project that other people could use if they quickly wanted to generate an assembler/disassembler for their own custom ISAs.

Issue #303, opened 2023-09-24
Untangle ISA-specific information from the assembler framework.


mortbopet:
Currently, defining an ISA and defining the assembler for said ISA is done in two different places, e.g.:
- https://github.com/mortbopet/Ripes/tree/master/src/isa
- RV-related files in https://github.com/mortbopet/Ripes/tree/master/src/assembler

Ideally, all information regarding operations themselves - such as opcodes, fields, registers, etc.. would not be inherently tied to the assembler, but instead part of the ISA information. The assembler framework would then query the operations of a given instruction set, and generate the assembler based on said information. This is _almost_ what's already done, but with the assembler being too strongly coupled with the instruction definitions.

If we manage to move instruction definitions to ISA files, meaning that the ISA files will contain **all** information about the ISA, we can start to query this more reliably in various different places in Ripes. One improvement that this refactor would allow is to avoid duplicating ISA information in the processor model decoders: https://github.com/mortbopet/Ripes/blob/master/src/processors/RISC-V/rv_decode.h - which instead should reference the ISA files.

Currrently, instructions are defined dynamically (e.g. https://github.com/mortbopet/Ripes/blob/master/src/assembler/rv_c_ext.h#L319C11-L323), meaning that all information that we specify when defining an instruction must lookup some runtime object to acquire the values of interest (opcode bitfields etc..). However, from the point of view of writing e.g. the RISC-V decoder model, we only need static information - hence there would be quite a bit of overhead in performing a bunch of dynamic, runtime lookups. As such, solving this issue should also consider whether there is a way to make instruction information (specifically bitfields) available at compile time.

PR #314, opened 2023-10-06
Make ISA definitions compile-time defined

raccog:
Draft branch for #306 and #311.


raccog:
Rewriting this branch from scratch as #315.

I'm going to also work on #303 and #307 within this branch. Figured since I am redefining a lot of types anyways, I might as well move them to the ISA library at the same time.

PR #320, opened 2023-11-14
Split register info class into multiple classes; one per register file

raccog:
Create `RegInfoInterface` as an abstract class to interface into a single register file. Return a vector of all the `RegInfoInterface`s implemented by an ISA.

Also refactored any code that indexes a register so that the register's file type is used in the index.

This includes the command line option `--reginit`. It now requires the register file type to be specified for each initialized register set. It can be used multiple times to initialize multiple register files. Example:

```bash
./Ripes --reginit gpr:1=1,20=0xdeadbeef --reginit csr:1=0xf
```


mortbopet:
I understand the intention of the change, but I'm not sure think that (as far as i understand) essentially merging multiple register infos in e.g. `RegInfoSet` is a proper representation. Each register file is distinct, and i think all places which care about providing a representation of the register files should represent each one as distinct, and not lump them together. 

I'd be more in favor of `ISAInfoBase` returning a list of `RegInfoInterface`, which each UI element then can use to provide distinct views of each register file: `std::vector<RegInfoInterface*> ISAInfoBase::getRegisterInfos()`

raccog:
> I'd be more in favor of `ISAInfoBase` returning a list of `RegInfoInterface`, which each UI element then can use to provide distinct views of each register file: `std::vector<RegInfoInterface*> ISAInfoBase::getRegisterInfos()`

@mortbopet  This sounds good to me since returning a `std::vector<RegInfoInterface*>` would serve the same purpose as `RegInfoSet` by providing a way to iterate over each available register from each register file that an ISA implements.

src/cli/clioptions.cpp
```diff
@@ -136,44 +137,70 @@ bool parseCLIOptions(QCommandLineParser &parser, QString &errorMessage,
...
+          regVal = decodeRadixValue(vstr, Radix::Signed, &ok);
+
+        if (!ok) {
+          errorMessage =
+              "Invalid register value '" + vstr + "' specified (--reginit).";
+          return false;
+        }
+
+        RegisterFileType regFileType;
+        if (regFile.startsWith("gpr", Qt::CaseInsensitive)) {
```
raccog, line 180: @mortbopet I had to change the `--reginit` command line parameter to accept a register file to index into. Do you think that the names `gpr`/`fpr`/`csr` are fine for specifying which register file to initialize? I feel like the set of register file types should be different for each ISA. For example, I don't think all ISAs have a CSR file.

What do you think about having a different `RegisterFileType` enum for each ISA?

mortbopet: How about extending `RegFileInfoInterface` with `RegFileInfoInterface::regFileName()`. Then, the CLI would iterate over the register file names exposed by the ISA, show it if there's a match, and if not, list the possible register file identifiers.

This would then also remove the need for the `RegisterFileType` since the whole point of that is to provide unique identifiers for each register file... might as well move that over to make the registerfile names load bearing 🤔.

raccog: I think that would work well, thanks!

src/isa/isainfo.h
```diff
@@ -47,28 +44,46 @@ struct RegInfoBase {
...
   virtual ISA isaID() const = 0;
+  virtual const RegInfoMap &regInfoMap() const = 0;
 
-  std::optional<const RegInfoBase *>
-  regInfo(RegisterFileType regFileType = RegisterFileType::GPR) const {
-    if (auto match = m_regInfos.find(regFileType); match != m_regInfos.end()) {
-      return m_regInfos.at(regFileType).get();
-    } else {
-      return {};
+  const std::set<std::string_view> regFileNames() const {
```
mortbopet, line 65: Why return a copied value as const?

src/isa/isainfo.h
```diff
@@ -47,28 +44,46 @@ struct RegInfoBase {
...
+    return regVec;
+  }
+  std::optional<const RegFileInfoInterface *>
+  regInfo(const std::string_view &regFileName) const {
+    if (auto match = regInfoMap().find(regFileName);
+        match != regInfoMap().end()) {
+      return {regInfoMap().at(regFileName).get()};
+    } else {
+      return {};
     }
```
mortbopet, line 86: Nit: sometimes it's considered better style to not have returns in else statements (https://clang.llvm.org/extra/clang-tidy/checks/readability/else-after-return.html).

src/registerinitializationwidget.cpp
```diff
@@ -27,23 +27,38 @@ void RegisterSelectionComboBox::showPopup() {
   const auto &procisa = ProcessorRegistry::getAvailableProcessors()
                             .at(m_parent->m_currentID)
                             ->isaInfo();
-  const auto regInfo = procisa.isa->regInfo().value();
+  const auto isa = procisa.isa;
```
mortbopet, line 30: nit: `const auto*`? (if we had clang-tidy checks enabled, it'd probably complain that you're auto'ing a pointer here).

src/registermodel.cpp
```diff
@@ -25,7 +25,12 @@ std::vector<VInt> RegisterModel::gatherRegisterValues() {
 int RegisterModel::columnCount(const QModelIndex &) const { return NColumns; }
 
 int RegisterModel::rowCount(const QModelIndex &) const {
-  return ProcessorHandler::currentISA()->regCnt();
+  if (auto regInfo = ProcessorHandler::currentISA()->regInfo(m_rft);
+      regInfo.has_value()) {
+    return (*regInfo)->regCnt();
+  } else {
+    return 0;
+  }
```
mortbopet, line 33: As above.

src/registermodel.cpp
```diff
@@ -129,15 +134,30 @@ void RegisterModel::setRadix(Ripes::Radix r) {
 }
 
 QVariant RegisterModel::nameData(unsigned idx) const {
-  return ProcessorHandler::currentISA()->regInfo().value()->regName(idx);
+  if (auto regInfo = ProcessorHandler::currentISA()->regInfo(m_rft);
+      regInfo.has_value()) {
+    return (*regInfo)->regName(idx);
+  } else {
+    return QVariant();
+  }
```
mortbopet, line 142: As above.

test/tst_cosimulate.cpp
```diff
@@ -97,23 +98,29 @@ private slots:
...
-      ProcessorHandler::get()->getProcessor()->implementsISA()->syscallReg());
-
-  /// @todo: Generalize this by having ISA report exit syscall codes
-  if (status == RVABI::SysCall::Exit || status == RVABI::SysCall::Exit2) {
-    m_stop = true;
+  if (auto reg = ProcessorHandler::get()
+                     ->getProcessor()
+                     ->implementsISA()
+                     ->syscallReg();
+      reg.has_value()) {
```
mortbopet, line 105: Nit: Generally, if-pyramids should be avoided; allows for easier to read code (and more optimized due to shorter branch distances). Instead, inverted if conditions and early returns are preferred. 

That is, this function could be rewritten as:
```c++
void tst_Cosimulate::trapHandler() {
  auto reg = ProcessorHandler::get()
                      ->getProcessor()
                      ->implementsISA()
                      ->syscallReg();
  if(!reg.has_value())
    return;
  
  unsigned status = ...
}
```

Also... should this be an assert? This would be a pretty wonky test if there wasn't a syscall register available x).

mortbopet (approved):
LGTM after these things are addressed.

src/cli/clioptions.cpp
```diff
@@ -136,44 +139,91 @@ bool parseCLIOptions(QCommandLineParser &parser, QString &errorMessage,
...
+        }
+
+        auto &vstr = regInitParts[1];
+        VInt regVal;
+        if (vstr.startsWith("0x"))
+          regVal = decodeRadixValue(vstr, Radix::Hex, &ok);
+        else if (vstr.startsWith("0b"))
+          regVal = decodeRadixValue(vstr, Radix::Binary, &ok);
+        else
+          regVal = decodeRadixValue(vstr, Radix::Signed, &ok);
```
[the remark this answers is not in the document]
raccog: Do you think that the `startsWith` comparisons should be case insensitive? So that (using hex as an example) `0X` or `0x` could be used as a prefix.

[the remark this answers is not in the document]
raccog: Never mind, I'm going to use the regex that are already defined in `radix.h` to decode the value, so `startsWith` won't actually be used.

src/isa/isainfo.h
raccog: I think that's leftover from my plan of returning a const reference 😅

src/isa/isainfo.h
raccog: Interesting, thanks for the tip!

Issue #275, opened 2023-04-21
Premature ECALL execution


Karazynol:
Using the `factorial.s` assembly example code, I noticed the `ecall` (here for printString) being output to the console before the command should have been able to run. 

As can be seen in the picture below I was using the Single-Cycle Prozessor and the `addi x17 x0 4` command has just finished when there's already a prompt to be seen. One can see the command being executed in the `Processor` View when continuing another step but apart from that virtually nothing else happens.
![grafik](https://user-images.githubusercontent.com/83511831/233684545-cee50929-2b6b-43f9-aa3a-7e80bd0faf57.png)

I believe sth similar happens for the 5-Stage Processor as well, in this case the prompt can be seen while `ecall` is still in the Execute stage
![grafik](https://user-images.githubusercontent.com/83511831/233692025-0d3a94c0-b092-4449-9793-99db1e66595b.png)

The same problem occurs when reverse stepping (undo) but I think this could be related to #57.

Issue #233, opened 2022-10-10
Branching before Forwarding Unit in 5-Stage without Hazard Detection


MicahDoo:
Hi,
I noticed in the **5-Stage without Hazard Detection Processor**, the outputs from Register Read are pushed directly into the Branch Comparator without going through the Forwarding multiplexers.

That means whatever instructions involve the Branch Comparator will not get the benefit of forwarding. Am I right?

I tested it using the following code:

With `beq`, two stalls are needed after an `addi`:
```asm
.data
branch_str:
    .string "Branched.\n"
not_branch_str:
    .string "Not branched.\n"
    

.text

main:
    
    # a0 set to 10
    addi a0, zero, 10
    nop
    
    # a0 set to 0
    addi a0, zero, 0
    
    # stalls (less than 2 nops => error (no branch))
    # nop
    # nop
    
    # if a0 = 0, branch => if no error, should branch
    beq a0, x0, branch
    
not_branch:
    la a0, not_branch_str
    li a7, 4
    nop
    nop
    nop
    ecall
    
    j exit
        
branch:
    la a0, branch_str
    li a7, 4
    nop
    nop
    nop
    ecall
    
exit:
    
    # Exit
    li a7, 10
    nop
    nop
    nop
    ecall
```
With `addi`, no stalls are needed:
```asm
.data
branch_str:
    .string "Branched.\n"
not_branch_str:
    .string "Not branched.\n"
    

.text

main:
    
    # a0 set to 10
    addi a0, zero, 10
    nop
    nop
    
    # a0 set to 0
    addi a0, zero, 0
    
    # zero stalls
    
    # no hazard: a0 = 0 + 3 = 3 / hazard: a0 = 10 + 3 = 13
    addi a0, a0, 3
    
    li a7, 1
    nop
    nop
    nop
    ecall
    

exit:
    
    # Exit
    li a7, 10
    nop
    nop
    nop
    ecall
```

mortbopet:
Hi @MicahDoo 

So is the issue here not that there is a RAW hazard on the following?
```asm
    addi a0, zero, 0
    # if a0 = 0, branch => if no error, should branch
    beq a0, x0, branch
```
It is exactly such a situation which a hazard unit is supposed to detect and then dynamically stall the pipeline to realize the 2-cycle bubble (as you indicate with the two `nop` instructions).

[...]

PR #325, opened 2023-12-05
Fix ecall support for read-only file operations

### LINKED ISSUE #316: Open ecall does not support read only mode
It is not possible to open a file in read only mode by passing `0` as `flags` parameter. The file will have the state `QIODevice::NotOpen` instead.

This seems to be related to a bug in the implementation of the ecall. Parsing the flags does not work for a `0` flag, as `flags & O_RDONLY` always evaluates to false since `O_RDONLY` equals `0`:  
<https://github.com/mortbopet/Ripes/blob/27fcc58c02b26e84156ef588d42fb36d8ada4543/src/syscall/systemio.h#L149-L154>

The only way to open a file for reading is using the read-write mode and then seeking back to position 0 in the file.

### Not working code using `O_RDONLY`

```s
.data

filename: .string "/tmp/test.txt"

.text

la a0 filename # a0: Pointer to filename
li a1 0 # Readonly flag
li a7 1024 # "Open" ecall
ecall # Returns: File descriptor

li a1 0x2000 # a1: buffer
li a2 100 # a2: max. buffer size
li a7 63 # "Read" ecall
ecall

li a0 0x2000
li a7 4 # "Print String" ecall
ecall
```

### Workaround using `O_RDWR` and seek

```s
.data

filename: .string "/tmp/test.txt"

.text

la a0 filename # a0: Pointer to filename
li a1 2 # Read-Write flag
li a7 1024 # "Open" ecall
ecall # Returns: File descriptor
mv s0 a0 # Save fd, since it will be overwritten by return values

nop # file descriptor stays in a0
li a1 0 # Offset 0
li a2 0 # From the beginning
li a7 62 # "Seek" ecall
ecall

mv a0 s0 # Get back file descriptor
li a1 0x2000 # a1: buffer
li a2 100 # a2: max. buffer size
li a7 63 # "Read" ecall
ecall

li a0 0x2000
li a7 4 # "Print String" ecall
ecall
```

raccog:
When working with file operations, the ecall instruction would not recognize the `O_RDONLY` flag. It would also accidentally use the `QIODevice::Append` flag in read-only mode, which would then silently fail and return the wrong characters. This fixes the flags so that read-only files return the correct characters (fixes #316).

Thanks to @lukasrad02 for reporting!

### DIFF
--- a/src/processorhandler.h
+++ b/src/processorhandler.h
@@ -62,9 +62,14 @@ public:
     return get()->_fullISA();
   }
 
-  /// Returns a reference to the system call manager.
+  /// Returns a const reference to the system call manager.
   static const SyscallManager &getSyscallManager() {
     return get()->_getSyscallManager();
+  }
+
+  /// Returns a non-const reference to the system call manager.
+  static SyscallManager &getSyscallManagerNonConst() {
+    return get()->_getSyscallManagerNonConst();
   }
 
   /// Sets the program p as the currently instantiated program.
@@ -294,6 +299,9 @@ private:
     return m_currentProcessor->fullISA();
   }
   const SyscallManager &_getSyscallManager() const { return *m_syscallManager; }
+  SyscallManager &_getSyscallManagerNonConst() const {
+    return *m_syscallManager;
+  }
   void _loadProcessorToWidget(vsrtl::VSRTLWidget *widget,
                               bool doPlaceAndRoute = false);
   void _selectProcessor(
--- a/src/statusmanager.h
+++ b/src/statusmanager.h
@@ -12,15 +12,17 @@ namespace Ripes {
 
 /**
  * @brief postToGUIThread
- * Schedules the execution of @param fun in the GUI thread.
+ * Schedules the execution of @param fun in the GUI thread if it exists.
  * @param connection type.
  */
 template <typename F>
 static void postToGUIThread(F &&fun,
                             Qt::ConnectionType type = Qt::QueuedConnection) {
-  auto *obj = QAbstractEventDispatcher::instance(qApp->thread());
-  Q_ASSERT(obj);
-  QMetaObject::invokeMethod(obj, std::forward<F>(fun), type);
+  if (qApp) {
+    auto *obj = QAbstractEventDispatcher::instance(qApp->thread());
+    Q_ASSERT(obj);
+    QMetaObject::invokeMethod(obj, std::forward<F>(fun), type);
+  }
 }
 
 class StatusEmitter : public QObject {
--- a/src/syscall/file.h
+++ b/src/syscall/file.h
@@ -14,10 +14,11 @@ class OpenSyscall : public BaseSyscall {
 
 public:
   OpenSyscall()
-      : BaseSyscall("Open", "Opens a file from a path",
-                    {{0, "Pointer to null terminated string for the path"},
-                     {1, "flags"}},
-                    {{0, "the file decriptor or -1 if an error occurred"}}) {}
+      : BaseSyscall(
+            "Open", "Opens a file from a path",
+            {{0, "Pointer to null terminated string for the path"},
+             {1, "flags" /* TODO(raccog): Add descriptions for each flag */}},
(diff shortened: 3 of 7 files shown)


raccog:
@lukasrad02 Thank you so much for continuing to submit well-written bug reports! Your descriptions have made it much easier to work on fixing these bugs. 👍 

> In general, it might be worth to consider changing the flags so that each flag has exactly one bit set to 1, so that simple bit operations work. This would – of course – not be backwards compatible, but if I did not miss anything in the commit history, the zero flag bug exists from the first release of the file I/O feature, hence backwards compatibility might not be necessary.

@mortbopet Gets the final say in this, but I think that we aren't too worried about keeping backwards compatibility for most of the internal structure of Ripes (the filesystem-cached settings are one exception to this, but these flags aren't used there). So, this sounds like a good solution to me.

> If you prefer to keep this PR small, I could also create a separate issue for the `writeToFile()` bug and/or the discussion on backwards compatibility.

No worries. Unless @mortbopet thinks that a separate discussion on backwards compatibility is necessary, we can keep this in a single issue/PR.

mortbopet:
Awesome - I'm not too concerned about backwards compatability; but you bring a good point @raccog that there should probably be some mechanism to invalidate cached program state from prior versions.

> One bug occurs when you clock and then undo an ecall instruction. This causes a SIGABRT, possibly from a failed host machine syscall because this is the disassembled code that causes the SIGABRT:

It's expected behavior that one cannot undo past an ecall and expect that whatever sideeffects that ecall had would have been undone. It _is_ technically possible, but i think it's a bit silly since it's a massive pain for people to have to implement undo-ing of all ecalls.
However, there should obviously not be a crash when doing this, so that's definitely a bug.

raccog:
> It's expected behavior that one cannot undo past an ecall and expect that whatever sideeffects that ecall had would have been undone. It _is_ technically possible, but i think it's a bit silly since it's a massive pain for people to have to implement undo-ing of all ecalls. However, there should obviously not be a crash when doing this, so that's definitely a bug.

I'll submit this as an issue soon.

> @raccog can you add your assembly program add your program as a test in https://github.com/mortbopet/Ripes/blob/master/test/tst_riscv.cpp?

Sure!

@mortbopet What's your opinion on the effects caused from the ecall instruction seemingly executing at the wrong stage/clock cycle?

>>The ecall instruction seems to be executed during the fetch stage instead of the execute stage. In the single-cycle processors, this causes the ecall instruction to be executed during the previous instruction's clock cycle instead of during the following clock cycle. On multi-cycle processors without hazard detection units, this causes the error `Unknown system call in register 'a7': 0` to appear. This is just what I think is happening from the testing I've done, I have not gone into the VSRTL code to verify my assumptions, yet.

I was wondering if you had any guesses as to why this might be happening. I haven't had the chance to look deeply at the VSRTL code yet to check my assumptions.

It only really seems like a problem in processors without hazard detection. I see similar bug reports in #275 and #233. In #233, you stated:

>The pipelined processors without hazard units are not supposed to execute a program correctly. They are meant as intermediate design points to be able to illustrate why hazard detection and resolution is required for a pipeline.
It should probably be more clear in the program/processor selection dialog that these processors aren't guaranteed to produce correct results..!

So, is the `Unknown system call in register 'a7': 0` error message considered normal for processors without hazard detection? Do you think it might be beneficial (or possible) to tell the user in this message that the error was caused by the lack of hazard detection in the processor? Or maybe write as a note in the processor's description that normal ecall usage is likely to fail?

mortbopet:
> So, is the Unknown system call in register 'a7': 0 error message considered normal for processors without hazard detection? Do you think it might be beneficial (or possible) to tell the user in this message that the error was caused by the lack of hazard detection in the processor? Or maybe write as a note in the processor's description that normal ecall usage is likely to fail?

I think it should be made clear(er) in the processor description. I'm unsure whether it's a good idea to add it to the error message, since it'll require some fairly processor specific logic (when's it a hazard related issue vs. actual unknown system call?) which (as you know) i like to keep away from Ripes (i.e. Ripes is a generic front-end for ISAs and their processor implementations).

Issue #334, opened 2023-12-24
Add Cache Replacement Policy


tinhanho:
Hi @mortbopet, 
I find that contributing to the cache replacement policy is a good way to work with. The replacement policy which possesses from now on are Random and LRU. Intuitively, FIFO can be added and it has been done through few lines of code. What I am doing now is trying to add policy like least frequently used. However, I cannot find out where the code is to expand the slot just like LRU doing. For example, when 2 ways cache used, you click Repl. policy LRU and the LRU slots show up. 
![image](https://github.com/mortbopet/Ripes/assets/67796326/0d375164-c13c-420b-80ee-6884767873aa)

I guess that it might be the UI problem but I do not know how to modify it. Could you help me?

HotMercury:
You can review line 440 in the incachegraphic.cpp file within issue #335, and you'll discover the relevant information there.
```
if (m_cache.getReplacementPolicy() == ReplPolicy::LRU && m_cache.getWays() > 1)
```

PR #335, opened 2023-12-24
Add Cache Replacement Policy

tinhanho:
About Issue [#334](https://github.com/mortbopet/Ripes/issues/334).

I think that we could add a FIFO mechanism to the cache policy. 

A new replacement policy, FIFO (First In, First Out), has been added to the existing enumeration. Additionally, a new counter has been introduced in the cachesim.h file. This counter plays a crucial role in cyclically determining which cache entry should be evicted.
```c
// line 19 in cachesim.h
enum ReplPolicy { Random, LRU, FIFO}; 
```

The code below handles the eviction process based on the FIFO replacement policy.
```c
//line 124 in cachesim.cpp
else if (m_replPolicy == ReplPolicy::FIFO){
    ew.first = CacheSim::counter;
    ew.second = &cacheLine[ew.first];
    CacheSim::counter += 1;
CacheSim::counter %= getWays();
}
```

### DIFF
--- a/src/cachesim/cachesim.cpp
+++ b/src/cachesim/cachesim.cpp
@@ -115,13 +115,19 @@ CacheSim::locateEvictionWay(const CacheTransaction &transaction) {
   std::pair<unsigned, CacheSim::CacheWay *> ew;
   ew.first = s_invalidIndex;
   ew.second = nullptr;
-
   // Locate a new way based on replacement policy.
   if (m_replPolicy == ReplPolicy::Random) {
     // Select a random way
     ew.first = std::rand() % getWays();
     ew.second = &cacheLine[ew.first];
-  } else if (m_replPolicy == ReplPolicy::LRU) {
+  } 
+    else if (m_replPolicy == ReplPolicy::FIFO){
+        ew.first = CacheSim::counter;
+        ew.second = &cacheLine[ew.first];
+        CacheSim::counter += 1;
+	CacheSim::counter %= getWays();
+   }
+    else if (m_replPolicy == ReplPolicy::LRU) {
     if (getWays() == 1) {
       // Nothing to do if we are in LRU and only have 1 set.
       ew.first = 0;
--- a/src/cachesim/cachesim.h
+++ b/src/cachesim/cachesim.h
@@ -16,7 +16,7 @@ class CacheSim;
 
 enum WriteAllocPolicy { WriteAllocate, NoWriteAllocate };
 enum WritePolicy { WriteThrough, WriteBack };
-enum ReplPolicy { Random, LRU };
+enum ReplPolicy { Random, LRU, FIFO};
 
 struct CachePreset {
   QString name;
@@ -91,7 +91,7 @@ class CacheSim : public CacheInterface {
   Q_OBJECT
 public:
   static constexpr unsigned s_invalidIndex = static_cast<unsigned>(-1);
-
+  int counter = 0;
   struct CacheSize {
     unsigned bits = 0;
     std::vector<QString> components;
@@ -329,7 +329,7 @@ private:
 };
 
 const static std::map<ReplPolicy, QString> s_cacheReplPolicyStrings{
-    {ReplPolicy::Random, "Random"}, {ReplPolicy::LRU, "LRU"}};
+    {ReplPolicy::Random, "Random"}, {ReplPolicy::LRU, "LRU"}, {ReplPolicy::FIFO, "FIFO"}};
 const static std::map<WriteAllocPolicy, QString> s_cacheWriteAllocateStrings{
     {WriteAllocPolicy::WriteAllocate, "Write allocate"},
     {WriteAllocPolicy::NoWriteAllocate, "No write allocate"}};


mortbopet:
Thank you for looking into adding a new cache replacement policy!

A few comments:
- As per the coding style currently used in Ripes, member variables should be prefixed with `m_` and not with the class name.
- I think the counter should be named better, e.g. `fifoIndexCounter`.
- (optional) do you think there is an accompanying visualization to this? i.e., for LRU, we also show the LRU bits. Could one imagine a similar column which indicates what way in the cache line is currently up for eviction as per. FIFO?

tinhanho:
Hi, @mortbopet.

Thanks for the reply. I had already renamed the counter and tried my best to follow the coding style. If there are any problem still, please let me know.

As for third comments, I rework all the framework to implement visualizing the FIFO bit. Now, there is no need about fifoIndexCounter. Instead, boolean fifoflag is presented. This flag is set under two circumstances,
1. When an invalid entry is selected, we set the fifoflag.
2. When all the entries are full and cache miss occurs, we need to choose a entry to evicted and we set the fifoflag.

When this flag is set, we shall add 1 to fifo bits if entry is valid. In this way, we'll find that when fifo bits equal to the way of the cache, that entry should be evicted.


```c 
//Line 167 in cachesim.cpp
if (it != cacheLine.end()) {
  ew.first = it->first;
  ew.second = &it->second;
  m_fifoflag = true;
}
if (ew.second == nullptr) {
  for (auto &way : cacheLine) {
    if (static_cast<long>(way.second.fifo) == getWays()){
      ew.first = way.first;
      ew.second = &way.second;
      m_fifoflag = true;
      break;
    }
  }
}
```

Furthermore, the undo part needs to be taken into consideration as well. If the policy is FIFO, we set the fifoflag again and we need to restore the oldway.

```c
//Line 442 in cachesim.cpp
if (!trace.transaction.isHit && getReplacementPolicy() == ReplPolicy::FIFO) {
  m_fifoflag = true;
  way = oldWay;
}
```

```c
//Line 82 in cachesim.cpp
if (getReplacementPolicy() == ReplPolicy::FIFO) {
  for(auto &set : line){
    if(set.second.valid && m_fifoflag) set.second.fifo--;
  }
  m_fifoflag = false;
  line[wayIdx].fifo = oldWay.fifo;
}
```

Demo video here,
https://github.com/mortbopet/Ripes/assets/67796326/b67aab30-17ee-4b3c-8fa7-ba56f905a282

Full code is in the branch FIFO of my fork,
https://github.com/tinhanho/Ripes/commit/6772d1abec6a6bff5e0e40374fd5922c67e3a427

Let me know if there are any problems of my think and implement.

mortbopet:
@tinhanho thank you for the further work - if you update the branch of this PR, it'll be a bit easier for me to review your code.

tinhanho:
@mortbopet I have already updated the branch 😉

src/cachesim/cachegraphic.cpp
```diff
@@ -98,27 +98,47 @@ void CacheGraphic::updateLineReplFields(unsigned lineIdx) {
...
+      const QString fifoText = QString::number(fifoVal);
+      way.second.fifo->setText(fifoText);
+
+      // LRU text might have changed; update LRU field position to center in
+      // column
+      const qreal y = lineIdx * m_lineHeight + way.first * m_setHeight;
+      const qreal x =
+          m_widthBeforefifo + m_fifoWidth / 2 - m_fm.horizontalAdvance(fifoText) / 2;
+      way.second.fifo->setPos(x, y);
+    }
```
mortbopet, line 139: LRU and FIFO code here is now essentially identical - please think about how to reduce code duplication here.

src/cachesim/cachegraphic.cpp
```diff
@@ -705,6 +744,14 @@ void CacheGraphic::initializeControlBits() {
             m_fm.horizontalAdvance(lruText) / 2;
         line[setIdx].lru = drawText(lruText, x, y);
       }
+      if (m_cache.getReplacementPolicy() == ReplPolicy::FIFO &&
+          m_cache.getWays() > 1) {
+        const QString fifoText = QString::number(0);
+        x = m_widthBeforefifo + m_fifoWidth / 2 -
+            m_fm.horizontalAdvance(fifoText) / 2;
```
mortbopet, line 751: Might be better to just do `m_fm.horizontalAdvance(QStringLiteral("0"))`

src/cachesim/cachesim.cpp
```diff
@@ -115,13 +140,48 @@ CacheSim::locateEvictionWay(const CacheTransaction &transaction) {
...
     // Select a random way
     ew.first = std::rand() % getWays();
     ew.second = &cacheLine[ew.first];
-  } else if (m_replPolicy == ReplPolicy::LRU) {
+  } 
+  else if (m_replPolicy == ReplPolicy::FIFO){
+//        ew.first = m_fifoIndexCounter;
+//        ew.second = &cacheLine[ew.first];
+//        m_fifoIndexCounter += 1;
+//	m_fifoIndexCounter %= getWays();
```
mortbopet, line 153: Remove commented out code

src/cachesim/cachesim.cpp
```diff
@@ -115,13 +140,48 @@ CacheSim::locateEvictionWay(const CacheTransaction &transaction) {
...
+   if (getWays() == 1) {
+      // Nothing to do if we are in LRU and only have 1 set.
```
mortbopet, line 155: Why a comment about LRU when we're in FIFO replacement policy?

PR #346, opened 2024-03-03
Add stack-heap collision check to BrkSyscall execution

federicovilla55:
Implemented additional checking in the `execute()` method of the brk syscall, defined in the `BrkSyscall` class, to ensure that memory allocation does not overflow stack space. The System Call has as its argument the new "program break", which is now compared with the current "stack pointer" before being set to avoid overlaps.
This commit resolves a @todo present in the code.


mortbopet:
Awesome! And then the obligatory question of - is it possible to add a unit test for this? :)

Issue #266, opened 2023-02-14
printf in cli mode


gits00:
Hello,
I can see that printf in code works in GUI mode but not in CLI mode. Is there any way it can be done in CLI mode or is it unsupported?
Thanks,

mortbopet:
This should ideally already be supported [here](https://github.com/mortbopet/Ripes/blob/565026f3b39c2151f3ba413e8192a7866b3e244a/src/cli/clirunner.cpp#L42-L45).

However, i do see that when running an example on Windows, nothing appears. Linux seems to be working as expected...

mortbopet:
fixed by https://github.com/mortbopet/Ripes/pull/348

PR #348, opened 2024-03-16
Add CLI Support for Executing ELF and C Files

federicovilla55:
In the Command Line Interface mode, although the possible execution file types are `c, asm, bin` (as shown by the command-line option `--t`), it is not possible to simulate `C` programs.

```
int Ripes::CLIRunner::processInput(): Assertion `false && "Command-line support for this source type is not yet implemented"' failed.
Aborted (core dumped)
```

This has been fixed with the possibility to execute *RISC-V* `ELF` files too. 

In the `clirunner.cpp` file I've added the logic to load three new source types: `C`, `InternalELF` and `ExternalELF`. 
For an ELF file first a check is done to be sure that the specified file is valid for the selected processor and then the file is loaded with the method `loadElfFile`. I moved this method to the `programutilities.cpp` file, following what was done for loading binary files with the method `loadFlatBinaryFile`. The function was moved so that it can be used both by CLI and GUI. 

To execute `C` programs first they are compiled using the already existing methods (such as `compileRaw()` from `ccmanager.cpp`), then they are converted as `ExternalELF` and loaded as mentioned above.

Since I've noticed that running CLI programs can sometimes have problems with large addresses (that cannot fit in an `int` variable), I've changed the type of the `byteAddress` variable to `long`. (Fixing #266)

### DIFF
--- a/src/cli/clirunner.cpp
+++ b/src/cli/clirunner.cpp
@@ -138,7 +138,7 @@ int CLIRunner::processInput() {
     QString fileContent = file.readAll();
     file.close();
     auto res = CCManager::get().compileRaw(fileContent, QString(),
-                                           /* disableGUI = */ false);
+                                           /* enableGUI = */ false);
     if (res.success) {
       m_options.src = res.outFile;
       m_options.srcType = SourceType::ExternalELF;
--- a/src/cli/clirunner.cpp
+++ b/src/cli/clirunner.cpp
@@ -1,5 +1,7 @@
 #include "clirunner.h"
+#include "ccmanager.h"
 #include "io/iomanager.h"
+#include "loaddialog.h"
 #include "processorhandler.h"
 #include "programutilities.h"
 #include "syscall/systemio.h"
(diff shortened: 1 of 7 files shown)


src/cli/clirunner.cpp
```diff
@@ -94,6 +96,59 @@ int CLIRunner::processInput() {
...
+    }
+    QFile file(m_options.src);
+    if (!file.open(QIODevice::ReadOnly)) {
+      error("Could not open file " + file.fileName());
+      return 1;
+    }
+    QString fileContent = file.readAll();
+    file.close();
+    auto res = CCManager::get().compileRaw(fileContent, QString(),
+                                           /* disableGUI = */ false);
```
mortbopet, line 141: `disableGUI` -> `showProgressDialog`. I get the intention here, but setting this value to `false` and naming it **disable**GUI implies that you're _enabling_ the gui.

src/cli/programutilities.h
```diff
@@ -8,4 +8,6 @@ namespace Ripes {
 QString loadFlatBinaryFile(Program &program, const QString &filepath,
                            unsigned long entryPoint, unsigned long loadAt);
 
+bool loadElfFile(Program &program, QFile &file);
```
mortbopet, line 11: Nit: now that you're technically re-declaring the function, i'd appreciate a comment describing the function (I'm sneakily trying to sneak in more documentation in the codebase 😉).

src/cli/programutilities.cpp
```diff
@@ -18,4 +23,121 @@ QString loadFlatBinaryFile(Program &program, const QString &filepath,
...
+      return nullptr;
+    *size_out = sec->get_size();
+    return sec->get_data();
+  }
+
+private:
+  elfio &reader;
+};
+
+std::shared_ptr<ELFIODwarfLoader> createDwarfLoader(elfio &reader) {
```
mortbopet, line 43: Nit: (i know this is my own code) - I think it's best practice to add `static` to `.cpp`-local function definitions to ensure that they are indeed private to the file - see https://llvm.org/docs/CodingStandards.html#anonymous-namespaces.

mortbopet (commented):
Sorry for the review delay; these are sensible changes. Some minor nits, and it should be good to go.

Re. the `int` fix - thank you for that; i'm sure there must be other places where narrowing issues like this are present (would probably be prudent to build Ripes with some `-Werror` to find these).

Issue #304, opened 2023-09-24
WASM: Reading console input does not work


mortbopet:
On https://ripes.me/ - to reproduce, run the `ConsoleReading.s` example and `Run` the processor.
Once prompted in the console, and the user types characters into the console, pressing `Enter` does not return control to the processor. The user is still able to pause running by reclicking `Run`, so not a critical error.

paramsiddharth:
Hi @mortbopet, just wanted to ask, is this issue planned to be fixed anytime soon?

mortbopet:
Not as far as I know - so it's free to be picked up in case you want to look into it!

federicovilla55:
Issue fixed in #360.

PR #360, opened 2024-05-02
Fix Issue: WASM, Reading console input does not work

federicovilla55:
In the current  WebAssembly (WASM) version of Ripes there's an issue preventing input reading from the console. This because pressing `Enter` does not return control to the processor, as stated in issue #304.

The problems stems from the behaviour of the `Console::keyPressEvent(QKeyEvent *e)` method, which requires a `QKeyEvent` parameter, that appears to behave differently in the binary executable compared to the WASM version. 
In the binary executable, when pressing Enter or Return, the `QKeyEvent` associated with the keys `Qt::Key_Return` or `Qt::Key_Enter` has a non empty text field (`e->text()`). However, in the WASM version that field is empty and therefore the check `if (!e->text().isEmpty())` always returns false when pressing Enter, resulting in the already reported error.
The same happens with the Backspace Key.

### Proposed fix

I propose altering the handling of these events: rather than mandating a non-empty text field in the `QKeyEvent`, I suggest to verify only the key associated with each command. Text inspection should only occur for the events where the key is not checked.

As stated in the [Qt documentation of the QKeyEvent Class of `key()`](https://doc.qt.io/qt-6/qkeyevent.html#key):
> Returns the code of the key that was pressed or released.
> See [Qt::Key](https://doc.qt.io/qt-6/qt.html#Key-enum) for the list of keyboard codes. These codes are independent of the underlying window system. 

The `key()` of an event is independent from the underlying system while the `text()` may not.

In the modified code, all types of `QKeyEvent` are processed using the same switch-case structure, as currently happens for the directional keys. When a key event is triggered, the program checks if the key corresponds to any of the arrow keys, the _Enter_ key, the _Return_ key or the _Backspace_ key.
If the key does not match any of these, the event is handled as it used to: first, it checks that the content of the key event is not empty, ensuring that modifier keys such as _Shift_, _Control_, or _Alt_ are not considered. Then, it adds the content of the event to the buffer and calls the `putData()` method.

### DIFF
--- a/src/console.cpp
+++ b/src/console.cpp
@@ -77,37 +77,34 @@ void Console::keyPressEvent(QKeyEvent *e) {
   case Qt::Key_Down:
     QPlainTextEdit::keyPressEvent(e);
     break;
+
+  case Qt::Key_Return:
+  case Qt::Key_Enter:
+    // Return is interpreted as \n instead of the default \r (\n)
+    m_buffer += "\n";
+
+    // Flush buffer to output
+    emit sendData(m_buffer.toLocal8Bit());
+    m_buffer.clear();
+
+    if (m_localEchoEnabled)
+      putData("\r");
+    break;
+
+  case Qt::Key_Backspace:
+    if (!m_buffer.isEmpty()) {
+      // Remove the last character from the buffer
+      m_buffer.chop(1);
+      if (m_localEchoEnabled)
+        backspace();
+    }
+    break;
+
   default:
     if (!e->text().isEmpty()) {
-      bool backspacedBuffer = false;
-      const QString text = e->text();
-      // Buffer managing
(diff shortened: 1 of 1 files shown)


mortbopet:
sorry for the delay; been on vacation for the past few weeks.

Thank you for looking into this - awesome that we're going to have a fix for this in WASM! have you tested this by running a local WASM build and (Linux | Windows) build to ensure that it works in both scenarios?

federicovilla55:
I've tested the PR with both local WASM and Linux/Windows builds, and it works in all cases. 
Let me know if there are any other scenarious I should check.

mortbopet:
@federicovilla55 sorry for the delay (feel free to ping me if i'm not replying)... awesome - we'll merge it :)

PR #384, opened 2024-12-31
Implement System Input Handling for CLI Execution

federicovilla55:
This commit implements a solution to handle system input within the Command-Line Interface (CLI), as in _Issue #353_.

Implemented changes:
- A new method, `SystemIO::setCLIInput(),` was added in the `systemio.h` file that sets the input stream to `STDIN`. This method is used only by the CLI, which doesn't support multi-thread execution and thus cannot rely on the same input handling methods as the GUI. The GUI has a dedicated thread for reading console input, while other threads manage the rest of the program.
- The `readFromFile()` method was updated to handle `STDIN` input differently. When `fd == STDIN`, it now reads one byte at a time using `InputStream.read(1)`. This continues until the specified number of bytes (`lengthRequested`) is reached or a newline (`'\n'`) character is encountered.
 
This change ensures that both the GUI and CLI can correctly handle `STDIN` input. Without reading one byte at a time, the CLI could not properly process inputs of a size different from the requested length, since there isn't a function like `putStdInData(const QByteArray &data)` in the CLI that pushes data into the `stdin` buffer object.
https://github.com/mortbopet/Ripes/blob/290c0e6be1bc280f9bdea8a74fe3e4c40145f025/src/syscall/systemio.h#L458-L463
For the GUI, STDIN is handled through the `QTextStream` object, which is linked in the `ConsoleWidget`:
https://github.com/mortbopet/Ripes/blob/290c0e6be1bc280f9bdea8a74fe3e4c40145f025/src/consolewidget.cpp#L23-L25

To integrate the new input handling in the CLI, a call to `SystemIO::setCLIInput()` is added in the `CLIRunner` constructor to set the input stream to `STDIN`:
```C++
CLIRunner::CLIRunner(const CLIModeOptions &options)
    : QObject(), m_options(options) {
  // to handle system input
  SystemIO::setCLIInput();
}
```

With these changes the `STDIN` input can be handled in both Ripes mode (GUI and CLI).

### DIFF
--- a/src/cli/clirunner.cpp
+++ b/src/cli/clirunner.cpp
@@ -11,8 +11,13 @@
 
 namespace Ripes {
 
-// An extended QVariant-to-string convertion method which handles a few special
-// cases.
+/**
+ * An extended QVariant-to-string convertion method which handles a special
+ * cases such as QVariantMap and QStringList.
+ * 
+ * @param v The QVariant to convert to a string.
+ * @returns A string representation of the QVariant.
+ */
 static QString qVariantToString(QVariant &v) {
   QString def = v.toString();
   if (!def.isEmpty())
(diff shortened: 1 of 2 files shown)


mortbopet:
https://github.com/mortbopet/Ripes/pull/385 should fix the failing CI test

mortbopet (approved):

Issue #386, opened 2025-01-05
Clipboard usage on WebAssembly


mortbopet:
Trying to use the clipboard in webassembly mode (https://ripes.me/) doesn't work. Most likely, it's due to some incorrect setup with the QT wasm build and/or the browser clipboard API.

Resources:
- https://www.qt.io/blog/qt-webassembly-clipboard

PR #389, opened 2025-01-30
Fix issue: Clipboard usage on WebAssembly

federicovilla55:
Fixed the `keyPressEvent` method in the `Console` class so that it is possible to copy/paste text from/to the clipboard with default handling shortcuts _Ctrl+C_ and _Ctrl+V_.
Pasted text is appended to `m_buffer` along with the rest of the typed input and displayed only if `m_localEchoEnabled` is `true.`
The standard method for _Qt_ Clipboard, `QGuiApplication::clipboard()`, is used.

This fixes issue #386.

### DIFF
--- a/.github/workflows/linux-release.yml
+++ b/.github/workflows/linux-release.yml
@@ -7,9 +7,6 @@ on:
 jobs:
   buildUbuntuXenial:
     runs-on: ubuntu-20.04
-    env:
-      CC: gcc-9
-      CXX: g++-9
 
     strategy:
       matrix:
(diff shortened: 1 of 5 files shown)


mortbopet:
Is this a full fix for #386 ? This looks to pertain only to the console. A probably more serious issue is the complete lack of modifier-related shortcuts for the editor (clipboard, but also selection-related (ctrl+A), ...).

This still looks like a good addition though, so i think we should get it in. I'll make a separate issue for the modifier-related issues.

Issue #347, opened 2024-03-05
Division by zero crash in assembler constant expressions


Tho-Del:
If an instruction that takes an immediate has a constant expression with denominator that is evaluated to 0, Ripes will crash from a division by zero evaluation error.

On windows, this error will require system registry edits to recover from, as the simulator loads and assembles the program before it can be edited, and the source code window is stored in the registry.
The relevant field is sourcecode in HKEY_CURRENT_USER\Softwware\Ripes\Ripes in the Registry Editor. 
Setting sourcecode empty or delete the relevant code enables Ripes to run without an instant crash again.

To reproduce:

paste 
```
la x0 0/0
```
or
```
start:
la x0 start/start
```
into Source code in Assembly mode. Upon assembling, Ripes will crash.

Cause:

Ripes/src/assembler/expreval.cpp evaluates assembler constant expression without checking for zero in denominator.
Also crashes with modulus % operator.

Tested in Ripes v2.2.6-41, windows 11

PR #390, opened 2025-01-31
Fix Issue: Division by zero crash in assembler constant expressions

federicovilla55:
Added checks in the expression evaluation (`evaluate`) method of `expreval.cpp` for division (`/`) and modulo (`%`) operations to prevent zero as a divisor.
Checks are performed by calculating and evaluating the _right-hand side value_ (`rhs_value`) of the binary operation. 
Used `std::runtime_error` to handle the case of division by zero so that the error is shown to the user directly in the written code in the text editor.

This fixes issue #347.


mortbopet:
Thank you for the fix!... and what an oversight on my part 🤦

PR #402, opened 2025-05-05
Accept labels that start with a dot.

ricferpas:
Ripes assembler gives an error for any identifier that starts with a dot. However, such syntax seems valid and is in fact routinely used by GCC. 
This patch changes this, making copying and pasting from Compiler Explorer more straightforward.


ricferpas:
Seems to display labels correctly both in the editor and the symbol explorer. I don't know if there is any other place where labels are used.

PR #403, opened 2025-05-05
Allow to set associativities of caches that are not power of 2.

ricferpas:
Ripes currently restricts the associativity of caches to powers of two. However, there is no technical reason to do so in hardware and there are caches in the real world whose associativity is not a power of two (6 and 12-way caches are common, and there are examples of 11-way caches (e.g.: https://en.wikichip.org/wiki/intel/microarchitectures/skylake_(server))).

This patch removes this restriction.


mortbopet:
LGTM, although the patch needs to be formatted.

PR #426, opened 2025-12-05
Fix shift amount masking for RV32

YerinAlexey:


mortbopet:
Just jogging my memory since it's been years since i visited this particular part of the ISA. So, we agree that this is not a problem for I-type instructions, correct? (since this masking is effectively applied [in the immediate unit](https://github.com/mortbopet/Ripes/blob/master/src/processors/RISC-V/rv_immediate.h#L61-L66). The problem is with R-type instructions, wherein the register value gets loaded directly, and the shift is applied directly (in the ALU unit).

Your fix then makes sense wrt. what is defined in the spec for the 32-bit ISA:
image
and for the 64-bit ISA:
image

This is pretty interesting; so basically, **all** of the open-source projects which use this particular test set, has this bug in it? (If you github serach for `"Verify that shifts only use bottom six bits"`, you'll see that many, many projects use these exact tests)...

Issue #344, opened 2024-02-08
M-extension issue (multiplication exceeds 64-bits)


AhmedAmrAbdellatif1:
Hello everyone,
I was trying to use `ripes` to verify my multiplier design unit, by trying to perform a calculation that exceed 64-bit to verify the implementation of `mulh` instruction.
So, i used the following assembly:
```
li x8, 11
slli x8, x8, 35
ori x8, x8, 254

li x9, 23
slli x9, x9, 35
ori x8, x8, 424

mul x10, x8, x9
mulh x11, x8, x9
mulhu x12, x8, x9
mulhsu x13, x8, x9
```

I'm trying to perform a multiplication between (`377957122558`) and (`790273982464`) the result should be (`2.986896804445448e+23`)
which in hexadecimal --> `0000000000003F40 00016E9000000000` ''calculated using MATLAB''

so the result of using `mul` should be `00016E9000000000`
and the result of using `mulh` should be `0000000000003F40`

> So, we can conclude that `ripes` calculate it wrong
![image](https://github.com/mortbopet/Ripes/assets/140100601/8e24e823-21a1-470f-8d23-ff96e2cc47a4)

PR #434, opened 2026-02-15
Fix incorrect MULH/MULHU/MULHSU calculation logic in 64-bit ALU

brian049:
This PR fixes a bug in the ALU implementation where `mulh`, `mulhu`, and `mulhsu` instructions were calculating incorrect results when running on a 64-bit processor model.

The original implementation shared the same logic for both 32-bit and 64-bit architectures (take `mulh` for example):

```cpp
    const auto result = static_cast<int64_t>(op1.sValue()) *
                        static_cast<int64_t>(op2.sValue());
    return VT_U(result >> 32);
```
For RV64, this logic is flawed because it truncates the 128-bit product (64-bit * 64-bit) into a 64-bit container and it shift by only 32 bits, which means returning bits [63:32] of the result instead of [127:64].

The commit updated `rv_alu.h` to check for `XLEN`. If `XLEN=64`, the ALU now uses `__int128_t` for calculation and shifts the result by 64 bits to correctly retrieve the high part of the product.

## Verification
I verified the fix using a custom assembly program provided in [issue#344](https://github.com/mortbopet/Ripes/issues/344), which performs multiplication that exceeds 64 bits.

<details>
<summary>Assembly Codes for Testing (from Issue #344)</summary>

```
li x8, 11
slli x8, x8, 35
ori x8, x8, 254

li x9, 23
slli x9, x9, 35
ori x8, x8, 424

mul x10, x8, x9
mulh x11, x8, x9
mulhu x12, x8, x9
mulhsu x13, x8, x9
```

</details>

Operand 1 (x8): `0x00000058000001fe`
Operand 2 (x9): `0x000000b800000000`
Expected High Part (x11): `0x0000000000003f40`

Before Fix:
Result in x11: `0x0000000000016E90` (Matches bits [63:32] of the product)

before

After Fix:
Result in x11: `0x0000000000003F40` (Matches the expected 128-bit high part)

after

### DIFF
--- a/src/processors/RISC-V/rv_alu.h
+++ b/src/processors/RISC-V/rv_alu.h
@@ -32,19 +32,40 @@ public:
       case ALUOp::MUL:
         return VT_U(op1.sValue() * op2.sValue());
       case ALUOp::MULH: {
-        const auto result = static_cast<int64_t>(op1.sValue()) *
-                            static_cast<int64_t>(op2.sValue());
-        return VT_U(result >> 32);
+        if constexpr (XLEN == 32) {
+          const auto result = static_cast<int64_t>(op1.sValue()) *
+                              static_cast<int64_t>(op2.sValue());
+          return VT_U(result >> 32);
+        } else {
+          const __int128_t result = static_cast<__int128_t>(op1.sValue()) *
+                                    static_cast<__int128_t>(op2.sValue());
+          return VT_U(result >> 64);
+        }
       }
       case ALUOp::MULHU: {
-        const auto result = static_cast<uint64_t>(op1.uValue()) *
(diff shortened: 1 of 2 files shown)


mortbopet:
Great! Mind adding (rewriting) the program that you've used to test into the [64-bit](https://github.com/mortbopet/Ripes/tree/master/test/riscv-tests-64) test suite?

brian049:
Hello @mortbopet , I've added the test to the 64-bit test suite as requested.

Additionally, I noticed the CI failed on the Windows build (due to MSVC not supporting __int128) and flagged some clang-format errors. I went ahead and included fixes for those issues as well to ensure all checks pass.

PR #447, opened 2026-04-26
Fix #444

### LINKED ISSUE #444: Incorrect EX stall in 5-stage pipeline
In the 5-stage pipeline implementation, when the offset in a `lw` happens to match the register number of the destination register in a previous `lw` instruction, the processor stalls the second `lw` in the `EX` stage for a single cycle, as if it were a load/use hazard.   The code below shows two 1 cycle stalls, which shouldn't be the case.  

I see this behavior in the ripes.me version and v2.2.6-77-g4981e31 binary. 

```
.data
A: .word 1, 2, 3, 4, 5, 6, 7, 8

.text
main:
    la   a0, A 

    # 1 cycle stall in EX for the lw t0
    lw   x28, 0(a0)
    lw   t0, 28(a0)

    # this also stalls in the EX stage
    lw   x12, 0(a0)
    lw   t0, 12(a0)

```

M-Enes:
Fixes bug stated in #444 by setting Rs2 to 0 when it is unavailable and also checking if destination register id is 0 in load-use hazard detection.


mortbopet:
Mind adding a test for it (which would fail without this fix)?

M-Enes:
Added tests. Also, I found that there is a similar case for Rs1. An example code:
```asm
.data
A: .word 5
.text
la a0, A
lw x14, 0(a0)
lui x6, 0x70
```
It is expected to last 8 cycles but it lasts 9 cycles (stalls 1 cycle). The root cause is that there is no Rs1 in U and J format instructions. So, bits 15-19 of the instruction are seen as Rs1 in the implementation. But, those bits are actually part of the immediate.
I am working on a fix for that.

Similar case exists for S and B format instructions (they don't have rd). But, it does not producing a functional bug. So, I think rd logic does not need to change.

M-Enes:
I have completed the fix and added tests. How does it look? Please let me know if anything needs to be changed.

mortbopet:
Awesome, thank you @M-Enes :).

PR #452, opened 2026-04-28
Use cache to speed up actions

M-Enes:
Adds ccache and sscache (for Windows builds) to speed up c++ compilation process. Also, uses github actions own cache mechanism to cache docker layers (alongside ccache mechanism).


mortbopet:
Awesome! Always nice to get some of these low hanging fruits in. This should really have been there to begin with...!

PR #457, opened 2026-05-09
fix the -Wdeprecated warnings intoduced by the multicycle processors

NLS-04:
fixed all -Wdeprecated warnings:
```
warning: implicit capture of ‘this’ via ‘[=]’ is deprecated in C++20 [-Wdeprecated]
```

by replacing `[=]` (which implicitely captures `this`) with `[this]` to be conform with C++20


mortbopet:
Thanks!

PR #461, opened 2026-05-10
Fix #460

### LINKED ISSUE #460: Tests does not run on macOS
Due to the `-executable` parameter for `find` command is not available on macOS, the tests don't work.
https://github.com/mortbopet/Ripes/blob/95d4b007f3e433da9578db39eea90d25268f7e3c/.github/workflows/scripts/run_tests.sh#L6

M-Enes:
Fixes #460. Due to the bug fixed in #459, the clang format action fails on this pull request. It could be ignored or merged after #459.

### DIFF
--- a/.github/workflows/scripts/run_tests.sh
+++ b/.github/workflows/scripts/run_tests.sh
@@ -3,7 +3,7 @@
 # Fail on failing QTest execution
 set -e
 
-find . -executable -type f -name "tst_*" | while read line; do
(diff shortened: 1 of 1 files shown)


mortbopet:
#459 has merged. As far as i understand, if you rebase this on main , it'll pass the format check, correct?

M-Enes:
Yes, it should pass the format check after rebase.

